PR #49, opened 2022-01-31
Fixed a bug causing inputs to be uncontrolled

code-person:
The bug is noticeable when the value of the input is not equal to what was entered by the user. For example, when certain characters are filtered out.

You can see in the screenshot below that there is a difference between the controlled and uncontrolled components:
image


franckgaudin:
Before accepting the PR, you must run the command `yarn changeset` to describe the changes made to the changelog

code-person:
> Before accepting the PR, you must run the command `yarn changeset` to describe the changes made to the changelog

Done ✅

syvuo (approved):

franckgaudin (approved):

PR #184, opened 2022-08-02
Feature/popover

franckgaudin:


syvuo (dismissed):

.storybook/styles.css
```diff
@@ -33,6 +42,32 @@ body,
   max-width: 100px;
 }
 
+.isb-trigger {
+  display: inline-block;
+  width: var(--space-2);
+  height: var(--space-2);
+  border-radius: 50%;
+  background-color: var(--electric-blue-500);
```
vicky-comeau, line 50: Maybe this doesn't apply to the storybook, but after reading the best practices, I think it said to use the theme variables and not the name of the actual color.

.storybook/styles.css
```diff
@@ -33,6 +42,32 @@ body,
...
+.isb-popover__tag {
+  border-radius: 50%;
+  width: var(--space-2);
+  height: var(--space-2);
+}
+
+.isb-popover__value {
+  font-size: var(--font-size-6);
+  font-weight: var(--font-weight-semi-bold);
+  color: var(--grey-800);
```
vicky-comeau, line 68: same as above

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,149 @@
...
+    title,
+    action,
+    isClosable = false,
+    dataTest,
+    ...rest
+  } = props;
+
+  const classes = cx('ids-popover__container', className);
+
+  const [show, setSow] = React.useState<boolean>(active);
```
vicky-comeau, line 56: Guessing this should be setShow and not setSow

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+  const onClose = (): void => {
+    setSow(false);
+  };
+
+  const popoverClasses = cx('ids-popover', popoverClassName);
+
+  const popoverStyle = {
+    ...styles.popper,
+    maxWidth: `${maxWidth}px`,
```
vicky-comeau, line 96: Again I'm just being picky, but does using rem instead of px not apply to maxWidth?

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,149 @@
...
+      <span
+        ref={setReferenceElement}
+        className={classes}
+        onClick={onClick}
+        role="button"
+        tabIndex={0}
+      >
+        {children}
+      </span>
+      {!show ? null : popover}
```
vicky-comeau, line 144: I think another way to write this could be:
{show && popover}
Since we are not showing anything otherwise.

packages/Tooltip/src/Tooltip.tsx
```diff
@@ -124,7 +124,7 @@ const Tooltip: React.FunctionComponent<TooltipProps> = (
       onMouseLeave={onMouseLeaveHandle}
     >
       {children}
-      {disabled ? null : tooltip}
+      {disabled || !show ? null : tooltip}
```
vicky-comeau, line 127: Same comment as above about using &&

utils/bootstrap/helpers/index.mjs
```diff
@@ -17,7 +17,7 @@ export async function checkRequiredProgramsExist(programs) {
 }
 
 export async function copyTemplate(directory) {
-  return await fs.copy('scripts/bootstrap/template', directory);
+  return await fs.copy('utils/bootstrap/template', directory);
```
vicky-comeau, line 20: I thought the docs said bootstrap was evil? haha maybe it's useful in some cases?

vicky-comeau (changes requested):

.storybook/styles.css
franckgaudin: yes in the front-end guide it says to use theme values. For the design system it's different, in each component we create local variables with their color value.
In the case of storybook, we use the variables directly, it's not a problem.

utils/bootstrap/helpers/index.mjs
franckgaudin: Bootstrap here is just a random name for the folder nothing to do with [Bootstrap](https://getbootstrap.com).
I'm not sure what bootstrap the doc is talking about 😉

packages/Popover/src/Popover.tsx
franckgaudin: That's a good point

vicky-comeau (approved):

syvuo (approved):

PR #194, opened 2022-08-23
Add Component Select

ofrogon:
Feat:
- Styles
  - Disables
  - Error
  - Compact
- Keyboard navigation
- `ReactNode` as options
- `Icon` in option


packages/Select/src/select.scss
```diff
@@ -0,0 +1,103 @@
...
+  &.ids-select--active,
+  &.active {
+    border-color: var(--ids-select-border-color-active);
+  }
+
+  &.ids-select--disabled {
+    color: var(--ids-select-color-disabled);
+    background-color: var(--ids-select-background-disabled);
+    border-color: var(--ids-select-border-color-disabled);
+    cursor: default;
```
vicky-comeau, line 97: for the disabled option, you put `cursor: not-allowed;` 
How comes you didn't do it for the actual disabled select?

packages/Select/src/select.scss
```diff
@@ -0,0 +1,103 @@
...
+  align-content: center;
+
+  &:focus,
+  &.focus {
+    outline: 0;
+    box-shadow: none;
+    border-color: var(--ids-select-border-color-active);
+  }
+
+  .ids-dropdown {
```
vicky-comeau, line 67: shouldn't the scss related to the dropdown be in the dropdown scss?

packages/Select/src/select.scss
```diff
@@ -0,0 +1,103 @@
...
+    position: absolute;
+    top: calc(var(--ids-select-height) - 1px);
+    left: -1px;
```
vicky-comeau, line 70: Could add a variable for these 1px in case this value is ever changed. It would also help with what this value means.

packages/Select/src/select.scss
```diff
@@ -0,0 +1,103 @@
...
+    display: flex;
+    flex-flow: column nowrap;
+  }
+
+  &.ids-select--compact {
+    font-size: var(--ids-select-font-size-compact);
+    line-height: var(--ids-select-line-height-compact);
+    height: var(--ids-select-height-compact);
+
+    .ids-dropdown {
```
vicky-comeau, line 82: Same as above comment

packages/Select/src/select.scss
```diff
@@ -0,0 +1,103 @@
...
+    position: absolute;
+    top: calc(var(--ids-select-height) - 1px);
+    left: -1px;
+    max-height: calc(var(--ids-select-height) * var(--ids-select-number-of-option-to-display));
+    overflow: auto;
+    display: flex;
+    flex-flow: column nowrap;
+  }
+
+  &.ids-select--compact {
```
vicky-comeau, line 77: do you need the extra class? if not, I think you can just write &--compact

packages/Select/src/SelectOption.tsx
```diff
@@ -0,0 +1,91 @@
...
+  React.useEffect(() => {
+    if (selected && selectOptionRef.current) {
+      selectOptionRef.current.scrollIntoView({
+        behavior: 'auto',
+        block: 'center',
+      });
+    }
+  }, [selected]);
+
+  const selectOptionClasses = cx('ids-select-option', className, {
```
franckgaudin, line 50: Here also you must use the BEM nomenclature. `ids-select__option`

website/example/Select.demo.js
```diff
@@ -0,0 +1,35 @@
+// import React from 'react';
```
vicky-comeau, line 1: Why did you comment this out? Is it because you're missing @igloo-ui/select inside next.config.js?

franckgaudin, line 1: Why is the code commented?

packages/Select/src/Select.tsx
```diff
@@ -0,0 +1,256 @@
+import * as React from 'react';
```
vicky-comeau, line 1: is it possible to combine all the select.tsx into one? It just seems to get kind of confusing looking at all these. It seems like maybe we're adding more code than we need to. Or maybe it's just me.

packages/Select/src/dropdown.scss
```diff
@@ -0,0 +1,50 @@
...
+  border-radius: var(--ids-dropdown-border-radius);
+  box-shadow: var(--ids-dropdown-box-shadow);
+  z-index: var(--ids-dropdown-zindex);
+
+  &--top {
+    transform: translateY(-100%);
+  }
+
+  &--right {
+    // TODO
```
vicky-comeau, line 30: Is this todo for another time or something you forgot?

vicky-comeau (changes requested):

packages/Select/src/select.scss
```diff
@@ -0,0 +1,103 @@
...
+  background-color: var(--ids-select-background);
+  border: 1px solid var(--ids-select-border-color);
+  border-radius: var(--ids-select-border-radius);
+  box-sizing: border-box;
+  appearance: textfield;
+  transition: border-color 200ms ease 0s;
+  margin: var(--ids-select-margin);
+  cursor: pointer;
+  user-select: none;
+  display: flex;
```
franckgaudin, line 54: Flex declarations have no effect here. Remove from line 54 to 58

packages/Select/src/select.scss
```diff
@@ -0,0 +1,103 @@
...
+  line-height: var(--ids-select-line-height);
+  height: var(--ids-select-height);
+  width: 100%;
+  padding: 0 var(--ids-select-padding);
+  position: relative;
+  background-color: var(--ids-select-background);
+  border: 1px solid var(--ids-select-border-color);
+  border-radius: var(--ids-select-border-radius);
+  box-sizing: border-box;
+  appearance: textfield;
```
franckgaudin, line 49: This property must be used on form elements otherwise it has on board effects for example in Safari

safari-render

packages/Select/src/select-value.scss
```diff
@@ -0,0 +1,51 @@
...
+
+  /* Placeholder */
+  --ids-dropdown-value-color-placeholder: #{tokens.$grey-400};
+}
+
+.ids-select-value {
+  display: flex;
+  flex-flow: row nowrap;
+  align-content: center;
+  gap: var(--ids-dropdown-value-gap);
```
franckgaudin, line 22: is not supported in IE11 :(

packages/Select/src/dropdown.scss
franckgaudin, line 30: He must miss something here 😀

packages/Select/src/SelectHeader.tsx
```diff
@@ -0,0 +1,38 @@
...
+) => {
+  const { children, disabled = false, isOpen = false, ...rest } = props;
+
+  const selectHeaderClasses = cx('ids-select-header', {
+    'ids-select-header--disabled': disabled,
+    'ids-select-header--active': isOpen,
+  });
+
+  return (
+    <div className={selectHeaderClasses} {...rest}>
```
franckgaudin, line 29: This div is not necessary, use `<></>` and put the logic of open, active, disabled at the parent

packages/Select/src/SelectHeader.tsx
```diff
@@ -0,0 +1,38 @@
...
+      {children}
+      <div className="ids-select-header__chevron">
```
franckgaudin, line 31: This div is not necessary, You can use the icon directly

packages/Select/src/SelectValue.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  const selectValueClasses = cx('ids-select-value', {
+    'ids-select-value--disabled': disabled,
+    'ids-select-value--compact': isCompact,
+    'ids-select-value--placeholder': !disabled && isPlaceholder,
+  });
+
+  return (
+    <div className={selectValueClasses} {...rest}>
+      {icon && <div className="ids-select-value__icon">{icon}</div>}
```
franckgaudin, line 39: This div is not necessary, you can use the icon directly

packages/Select/src/SelectValue.tsx
```diff
@@ -0,0 +1,45 @@
...
+      <div className="ids-select-value__label">{label}</div>
```
franckgaudin, line 40: This div is not useful

packages/Select/src/select.scss
ofrogon: When writing the `Select` classes I just forgot that we normally set `cursor: not-allowed;` do disabled components.

I will fix that oversight ✏

packages/Select/src/select.scss
ofrogon: I was not sure where to put that CSS, it's for specific `Select` behavior, but for the `Dropdown`.
I opted to add it to the `select.scss` instead of `dropdown.scss` to keep `Dropdown` as generic as possible.

website/example/Select.demo.js
ofrogon: I wasn't able to make this section work, I commented it with a plan of asking for help later and don't have the time to fix + forgot it after.

packages/Select/src/Select.tsx
ofrogon: I personally prefer to separate many components in many files (1 component by file).

packages/Select/src/dropdown.scss
ofrogon: It's a TODO for when/if the Dropdown bacame it's own component.
It's maybe overkill for now to have it now because it's trivial that it will be needed base on `Dropdown.tsx`.
image

packages/Select/src/select.scss
vicky-comeau: hm I suppose this makes sense.

packages/Select/src/dropdown.scss
vicky-comeau: oh I get it now :) makes sense

packages/Select/src/SelectHeader.tsx
```diff
@@ -0,0 +1,38 @@
...
+  /** True if the option list is displayed. */
+  isOpen?: boolean;
+}
+
+const SelectHeader: React.FunctionComponent<SelectHeaderProps> = (
+  props: SelectHeaderProps
+) => {
+  const { children, disabled = false, isOpen = false, ...rest } = props;
+
+  const selectHeaderClasses = cx('ids-select-header', {
```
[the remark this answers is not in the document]
ofrogon: Est-ce que je devrais aussi renommer le fichier en `SelectInput`? A moins que la différence entre le nom de la classe et le nom du component ne dérange pas.

website/example/Select.demo.js
ofrogon: Beacause I wasn't able to make it works.
I will need help from you for this.

packages/Select/src/select-value.scss
ofrogon: IE11 is dead since mid-june, did we still support it?

packages/Select/src/SelectHeader.tsx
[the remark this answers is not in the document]
franckgaudin: Tu peux le changer c'est jamais négatif d'avoir une continuité dans la nomenclature.

packages/Select/src/SelectHeader.tsx
ofrogon: Pour permettre de séparer ce qui est "Input" de ce qui est "Dropdown" au niveau du `Select` c'est bien pratique d'avoir un `<div>` qui englobe le tout.

vicky-comeau (approved):

packages/Select/src/Dropdown.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+    return transitions((styles) => (
+      <>
+        <div
+          role="button"
+          tabIndex={0}
+          className="ids-dropdown__overlay"
+          onClick={handleOnClose}
+        >
+          {null}
```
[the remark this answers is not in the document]
ofrogon: EsLint cry if I let a `div` empty (`<div></div>`) or if I set it as self-close (`<div/>`).

Using `{null}` fix the warning.

packages/Select/src/select-value.scss
ofrogon: Ok, I will change that 👍

franckgaudin (approved):

vicky-comeau (approved):

PR #195, opened 2022-08-23
Feat: Tabs component

vicky-comeau:
Created the tabs component


packages/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,142 @@
...
+    }
+  };
+
+  const renderTabButton = (
+    tab: TabInterface,
+    index: number
+  ): React.ReactNode => {
+    let button = (
+      <button
+        className="ids-tabs__tab-button"
```
franckgaudin, line 72: As we are in the block `tab` it is better to use this nomenclature `ids-tab__btn` for CSS. 
And to keep a consistency with the other components replace **button** by **btn**

packages/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,142 @@
...
+        onClick={() => handleOnClick(index)}
+      >
+        {tab.label}
+
+        {tab.notification && (
+          <TextBulletSolid
+            size="small"
+            className="ids-tabs__tab-icon ids-tabs__bullet-icon"
```
franckgaudin, line 80: idem here use `ids-tab__icon` `ids-tab__bullet`

franckgaudin (commented):
We must discuss the operation with React Router, the component must remain as generic as possible. It is preferable that it is to manage the user side.

packages/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,142 @@
...
+  className,
+  isInline = true,
+  tabs,
+}) => {
+  const classes = classNames('ids-tabs', className, {
+    'ids-tabs--inline': isInline,
+    'ids-tabs--heading': !isInline,
+  });
+
+  const [selectedTabIndex, setSelectedTabIndex] =
```
franckgaudin, line 55: Igloo components are unControlled components, for the moment we leave the state management for the user's. You can refer to the `Radio` or `Input` component.

packages/Tabs/README.md
```diff
@@ -0,0 +1,51 @@
...
+```bash
+yarn add @igloo-ui/tabs
+```
+
+## Usage
+
+Then to use the component in your code just import it!
+
+```jsx
+import Tabs, { TabPane } from '@igloo-ui/tabs';
```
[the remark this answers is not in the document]
vicky-comeau: Good catch, I missed that.

packages/Tabs/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "Vicky Comeau vicky.comeau@gsoft.com"
+  ],
+  "license": "Apache-2.0",
+  "peerDependencies": {
+    "react": ">= 16.8.6"
+  },
+  "files": [
+    "dist"
+  ],
+  "dependencies": {
```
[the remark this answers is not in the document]
vicky-comeau: right, I'll add it.

packages/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,142 @@
+import * as React from 'react';
+
+import './tabs.scss';
+
+import { useState } from 'react';
```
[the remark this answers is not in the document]
vicky-comeau: done :)

packages/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,142 @@
...
+import classNames from 'classnames';
```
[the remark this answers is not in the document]
vicky-comeau: right, I think I used that in the other component, I'll remember that, thanks.

packages/Tabs/src/Tabs.tsx
vicky-comeau: alright, thanks for the tip

packages/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,114 @@
...
+    } else {
+      tabItem = <span className="ids-tab__nav">{tabItem}</span>;
+    }
+
+    return (
+      <li
+        key={index.toString()}
+        className={cx('ids-tab', {
+          'ids-tab--active': index === selected,
+          'ids-tab--has-btn': isTypeString,
```
franckgaudin, line 92: This class is not used, I do not find it in the `.scss`. Idem for `ids-tab--no-btn`

website/example/Tabs.demo.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import Tabs from '@igloo-ui/tabs';
+
+const Example = () => {
```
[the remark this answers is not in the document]
vicky-comeau: Good catch again, fixed

packages/Tabs/src/Tabs.tsx
vicky-comeau: Removed the unused classes

franckgaudin (dismissed):

packages/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,112 @@
...
+  className,
+  isInline = true,
+  tabs,
+}) => {
+  const classes = cx('ids-tabs', className, {
+    'ids-tabs--inline': isInline,
+    'ids-tabs--heading': !isInline,
+  });
+
+  /* Handle the on click */
```
ofrogon, line 51: I don't think we need a comment here, the constant name already define what it is good enough.

packages/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const handleOnClick = (index: number): void => {
+    if (onSelectTab) {
+      onSelectTab(index);
+    }
+  };
+
+  /* Render the Tab */
```
ofrogon, line 58: I don't think we need a comment here, the constant name already define what it is good enough.

packages/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,112 @@
...
+    );
+  };
+
+  return (
+    <div className={classes} data-test={dataTest}>
+      <ul className="ids-tabs__list">
+        {tabs.map((tab, index) => renderTabItem(tab, index))}
+      </ul>
+
+      {/* show selcted tab by index */}
```
ofrogon, line 106: Typo in `selected` + if you need to comment what you done, maybe extract the logic in a method with a good verbose name.

packages/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,112 @@
...
+export interface TabInterface {
+  /** The tab pane contents of the tabs list */
+  children?: React.ReactNode;
+  /** Add an id attribute to the tab */
+  id?: string;
+  /** Add text for the tab */
+  label: React.ReactNode;
+  /** Add a notification indicator to the tab */
+  notification?: boolean;
+  /** Add a premium indicator to the tab */
```
ofrogon, line 19: In you documentation, you should decide if you write `Tab` or `tab` to have continuity between each other.

ofrogon (commented):
Some comments do remove, maybe a method to extract. Other than that, good job!

packages/Tabs/src/Tabs.tsx
vicky-comeau: Sometimes I like to put comments so it's easier to find certain things when scrolling through the component, but if you want I can remove them.

ofrogon: My thinking with comments is that it's always obsolete once you write it. You never know when someone else will change/refactor something that will make a comment tell a lie.

So overall, I prefer to set the less comment possible and only have one if it's impossible to easily do it directly in the code.

Some comments I think we can accept:
- For `REGEX`
- For complex `switch-case`
- Sometime, to define sections

franckgaudin (approved):

packages/Tabs/src/Tabs.tsx
vicky-comeau: makes sense. I have removed them.

packages/Tabs/src/Tabs.tsx
vicky-comeau: I ended up removing this comment.

packages/Tabs/src/Tabs.tsx
vicky-comeau: I decided to use lowercase T since it's proper English that way..

ofrogon (dismissed):

ofrogon (approved):

franckgaudin (approved):

PR #197, opened 2022-08-25
Fix: Tooltip component

vicky-comeau:
Changed the mouseleave event to use native js so it works with disabled elements as children.

### DIFF
--- a/packages/Tooltip/src/Tooltip.tsx
+++ b/packages/Tooltip/src/Tooltip.tsx
@@ -119,12 +119,27 @@ const Tooltip: React.FunctionComponent<TooltipProps> = (
     document.body
   );
 
+  React.useEffect(() => {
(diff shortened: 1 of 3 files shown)


packages/Tooltip/src/Tooltip.tsx
```diff
@@ -119,12 +119,27 @@ const Tooltip: React.FunctionComponent<TooltipProps> = (
     document.body
   );
 
+  React.useEffect(() => {
+    if (referenceElement) {
+      // The reaosn we use the native js event is because react has a bug where
```
franckgaudin, line 124:
```suggestion
      // The reason we use the native js event is because react has a bug where
```

franckgaudin (approved):

ofrogon (approved):

PR #199, opened 2022-08-29
Fix/issues with doc

franckgaudin:
In this PR here is what was done:

- Adding an example for `Portal`
- Adding an example for `Popover` and an image
- Fix of the opening of the modal in Storybook


website/styles/components/_example.scss
```diff
@@ -29,4 +29,82 @@
...
+  &__header {
+    display: flex;
+    align-items: flex-start;
+    justify-content: space-between;
+  }
+
+  &__profile {
+    display: flex;
+    align-items: center;
+    gap: rebase(1rem);
```
vicky-comeau, line 51: This is more of a question but what is rebase? I haven't seen this before.

franckgaudin: The rebase is a helper which transforms the values into base 10. 
The component is developed with tokens on base10 and In the doc we base ourselves on a base 8.

**Example**
Component: 1rem = 10px;
Doc: 1rem = 16px;

vicky-comeau (approved):

ofrogon (approved):

PR #207, opened 2022-09-09
feat(buttonGroup): add support for icon and text

franckgaudin:
Jira: #OV-39150

### DIFF
--- a/packages/ButtonGroup/src/ButtonGroup.stories.tsx
+++ b/packages/ButtonGroup/src/ButtonGroup.stories.tsx
@@ -137,3 +137,25 @@ export const WhitIcons = () => (
     </ButtonGroup>
   </Section>
 );
+
+export const WhitIconsAndText = () => (
+  <Section>
+    <ButtonGroup>
+      <ButtonItem active>
+        <UnorderedList /> List
+      </ButtonItem>
+      <ButtonItem>
+        <Tree /> Tree
+      </ButtonItem>
+    </ButtonGroup>
+
+    <ButtonGroup small>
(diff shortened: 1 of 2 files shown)


packages/ButtonGroup/src/button-group.scss
```diff
@@ -83,6 +88,10 @@
     margin-left: -1px;
   }
 
+  > svg {
+    margin-right: var(--ids-button-group-icon-spacing);
```
vicky-comeau, line 92: The margin right is always there, even when there is no text (except for compact ones). There should be no margin right if there is no text. It could maybe be written a different way. For example, put margin left on the text instead, but only when the svg is there. svg + .text {margin-right:}

packages/ButtonGroup/src/ButtonGroup.stories.tsx
```diff
@@ -101,39 +101,39 @@ export const Small = () => (
...
-      </ButtonItem>
-      <ButtonItem>
-        <Tree />
+      <ButtonItem icon={<UnorderedList />} />
+      <ButtonItem icon={<Tree />} />
+    </ButtonGroup>
+  </Section>
+);
+
+export const WhitIconsAndText = () => (
```
CarolineSabourin-Jean, line 125: With?

CarolineSabourin-Jean (approved):

CarolineSabourin-Jean (approved):

ofrogon (approved):

vicky-comeau (approved):

PR #212, opened 2022-09-15
feat: initial release of OptionButton

vicky-comeau:


packages/OptionButton/src/OptionButton.tsx
```diff
@@ -0,0 +1,141 @@
...
+        return null;
+    }
+  };
+
+  const classes = cx('ids-option-button', className, {
+    'ids-option-button--unchecked': unchecked,
+  });
+
+  return (
+    <span className={classes} data-test={dataTest}>
```
franckgaudin, line 118: in the css you declare `display: block;`it would be better in this case to use a **div**

packages/OptionButton/src/option-button.scss
```diff
@@ -0,0 +1,145 @@
...
+    font-family: var(--ids-option-button-font-family);
+    outline: none;
+    overflow: hidden;
+  }
+
+  &__radio:hover ~ &__label {
+    border-color: var(--ids-option-button-hover-border-color);
+  }
+
+  &__radio:focus ~ &__label {
```
franckgaudin, line 72: For the focus style use the property **shadow**and the `var(--focus)` because on safari the border radius are broken with the outline.

Capture d’écran, le 2022-09-16 à 12 13 49

packages/OptionButton/src/option-button.scss
```diff
@@ -0,0 +1,145 @@
...
+%disabled-icon-container {
+  background: var(--ids-option-button-icon-container-disabled-background);
+}
+
+.ids-option-button {
+  display: block;
+  position: relative;
+
+  &__radio {
+    opacity: 0;
```
franckgaudin, line 53: In the `Toggle` we used this technique to hide the radio button, if you want to keep a continuity

```
{
  height: 0;
  opacity: 0;
  position: absolute;
  width: 0;
}
```

packages/OptionButton/src/option-button.scss
```diff
@@ -0,0 +1,145 @@
...
+    min-width: 0;
+    padding: var(--ids-option-button-text-container-padding);
+  }
+
+  &__text {
+    color: var(--ids-option-button-text-color);
+    display: block;
+    font-size: var(--ids-option-button-text-font-size);
+    font-weight: var(--ids-option-button-text-font-weight);
+    overflow: hidden;
```
franckgaudin, line 132: you can group the declarations for the ellipsis in a class `%ellipsis` and repeat it in `&__text` and `&__desc`

packages/OptionButton/src/OptionButton.tsx
```diff
@@ -0,0 +1,141 @@
...
+  /** Add a description that is displayed below the text prop */
+  description?: string;
+  /** True if the option button should be disabled */
+  disabled?: boolean;
+  /** Add an icon the the beginning of the option button.
+   * This overrides the icon that comes with buttonType */
+  icon?: React.ReactNode;
+  /** Add an id to the radio button of the option button.
+   * This will also be used as the htmlFor attribute on the label */
+  id: string;
```
franckgaudin, line 35: On `Toggle` and `Radio` we have named this props `htmlFor` to keep a constancy can you change it ?

packages/OptionButton/src/OptionButton.tsx
```diff
@@ -0,0 +1,141 @@
...
+  className?: string;
+  /** Add a data-test tag for automated tests */
+  dataTest?: string;
+  /** Add a description that is displayed below the text prop */
+  description?: string;
+  /** True if the option button should be disabled */
+  disabled?: boolean;
+  /** Add an icon the the beginning of the option button.
+   * This overrides the icon that comes with buttonType */
+  icon?: React.ReactNode;
```
franckgaudin, line 32: Did you validate with Francis this case? Is it possible that this component is a specific icon?

packages/OptionButton/src/OptionButton.tsx
```diff
@@ -0,0 +1,141 @@
...
+  /** Add an id to the radio button of the option button.
+   * This will also be used as the htmlFor attribute on the label */
+  id: string;
+  /** Add an onChange event to the option button */
+  onChange?: (e: React.ChangeEvent<HTMLInputElement>) => void;
+  /** Added the main text to the option button */
+  text: string;
```
franckgaudin, line 39: I question myself **text** vs **children**

packages/OptionButton/src/OptionButton.tsx
```diff
@@ -0,0 +1,141 @@
...
+  /** Add a button type, which will come with its own icon.
+   * If adding an icon, this will be overridden */
+  buttonType?: ButtonType;
```
franckgaudin, line 42: here we can simply use `type`

packages/OptionButton/src/OptionButton.tsx
```diff
@@ -0,0 +1,141 @@
+import * as React from 'react';
+import cx from 'classnames';
+
+import {
+  TextIcon,
+  OptionScaleIcon,
+  MultipleChoiceIcon,
+  LikertIcon,
+  TextIconDisabled,
```
franckgaudin, line 9: for reduce the number of icons and the size of file you can manage the color via css. I put an example [here](https://codesandbox.io/s/icon-color-disabled-reikpt?file=/src/App.js).

packages/OptionButton/src/OptionButton.tsx
vicky-comeau: because I am extending a radio button, I could not use type. They are not the same

packages/OptionButton/src/OptionButton.tsx
vicky-comeau: yes, he said adding the icon prop was fine.

packages/OptionButton/src/OptionButton.tsx
vicky-comeau: hmm I think we only want them to enter a string and not html, right?

packages/OptionButton/src/OptionButton.tsx
franckgaudin: ok in this case leave it like that

packages/OptionButton/src/OptionButton.tsx
franckgaudin: You can always pass a string for the children. 

It's more than a question of implementation.
`<OptionButton buttonType='text" '>Text</OptionButton>`
   **VS** 
`<OptionButton buttonType="text" text="Text" />`

vicky-comeau: ok then I'll switch it to children. I like that to.

packages/OptionButton/src/OptionButton.tsx
```diff
@@ -0,0 +1,128 @@
...
+  onChange?: (e: React.ChangeEvent<HTMLInputElement>) => void;
+  /** True if the option button should have an unchecked appearance */
+  unchecked?: boolean;
+}
+
+const OptionButton: React.FunctionComponent<OptionButtonProps> = (
+  props: OptionButtonProps
+) => {
+  const {
+    buttonType,
```
franckgaudin, line 47: As the property is not required, I would default to the text type
```suggestion
    buttonType: 'text',
```

packages/OptionButton/src/OptionButton.tsx
```diff
@@ -0,0 +1,128 @@
...
+import {
+  TextIcon,
+  OptionScaleIcon,
+  MultipleChoiceIcon,
+  LikertIcon,
+} from './svgs';
+
+import './option-button.scss';
+
+export type ButtonType = 'Text' | 'OptionScale' | 'MultipleChoice' | 'Likert';
```
franckgaudin, line 13: For unions we use the camelCase
```suggestion
export type ButtonType = 'text' | 'optionScale' | 'multipleChoice' | 'likert';
```

packages/OptionButton/src/OptionButton.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const renderIcon = (): React.ReactNode => {
+    if (icon) {
+      return icon;
+    }
+    let disabledClass = '';
+    if (disabled || unchecked) {
+      disabledClass = 'disabled';
+    }
+    switch (buttonType) {
+      case 'Text':
```
franckgaudin, line 70: if the default `buttonType` is `text` remove this case and place on default return

franckgaudin (dismissed):

franckgaudin (dismissed):

packages/OptionButton/src/option-button.scss
```diff
@@ -0,0 +1,158 @@
...
+@use '~@igloo-ui/tokens/dist/fonts';
+
+:root {
+  /* Default */
+  --ids-option-button-border-color: #{tokens.$grey-400};
+  --ids-option-button-border-radius: #{tokens.$border-radius-sm};
+  --ids-option-button-font-family: #{tokens.$primary-font-family};
+
+  /* Focus */
+  --ids-option-button-focus-outline-color: #{tokens.$electric-blue-600};
```
ofrogon, line 11: `--ids-option-button-focus-outline-color` is not used

vicky-comeau: right, that was changed so I can remove that.

packages/OptionButton/src/option-button.scss
```diff
@@ -0,0 +1,158 @@
...
+  --ids-option-button-icon-container-height: 5.6rem;
+  --ids-option-button-icon-container-width: 6.2rem;
+  --ids-option-button-icon-background-disabled: #{tokens.$grey-300};
+
+  /* Text */
+  --ids-option-button-text-container-padding: #{tokens.$space-1} #{tokens.$space-3};
+  --ids-option-button-text-color: #{tokens.$grey-800};
+  --ids-option-button-text-font-size: #{tokens.$font-size-5};
+  --ids-option-button-text-font-weight: #{tokens.$font-weight-semi-bold};
+  --ids-option-button-text-line-height: #{tokens.$line-height-xl};
```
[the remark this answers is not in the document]
vicky-comeau: I added the variable but forgot to use it :/

packages/OptionButton/src/OptionButton.tsx
```diff
@@ -0,0 +1,126 @@
...
+  /** Add a data-test tag for automated tests */
+  dataTest?: string;
+  /** Add a description that is displayed below the text prop */
+  description?: string;
+  /** True if the option button should be disabled */
+  disabled?: boolean;
+  /** Add an htmlFor to the radio button of the option button.
+   * This will also be used as the htmlFor attribute on the label */
+  htmlFor: string;
+  /** Add an icon the the beginning of the option button.
```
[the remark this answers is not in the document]
vicky-comeau: good catch :P

franckgaudin (approved):

ofrogon (approved):

PR #217, opened 2022-09-20
Feat: form group

vicky-comeau:


packages/FormGroup/src/FormGroup.stories.tsx
```diff
@@ -0,0 +1,122 @@
+import React from 'react';
+
+import { ComponentMeta, ComponentStory } from '@storybook/react';
+
+import Section from '@components/section';
+import readme from '../README.md';
+
+import FormGroup from './FormGroup';
+import Input from '@igloo-ui/input';
```
franckgaudin, line 9: it's a small detail but place the module import in first time, a the second place the intern module at last the local file.

**Example:** 
```
import React from 'react';
import Input from '@igloo-ui/input';
...

import Section from '@components/section';
...

import readme from '../README.md';
import FormGroup from './FormGroup';

```

packages/FormGroup/src/FormGroup.stories.tsx
```diff
@@ -0,0 +1,122 @@
...
+  title: 'Components/FormGroup',
+  component: FormGroup,
+  parameters: {
+    description: readme,
+  },
+} as ComponentMeta<typeof FormGroup>;
+
+const Template: ComponentStory<typeof FormGroup> = (args) => (
+  <FormGroup {...args}>
+    <Input type="text" placeholder="John" error={true} />
```
franckgaudin, line 23: it's possible to link the `showError` with the input error ?

packages/Textarea/src/textarea.scss
```diff
@@ -48,6 +48,7 @@
     border: 1px solid var(--ids-textarea-border-color);
     border-radius: var(--ids-textarea-border-radius);
     color: var(--ids-textarea-color);
+    display: block;
```
franckgaudin, line 51: I don't understand why a `display: block` here?

vicky-comeau: otherwise we have extra space on the bottom of the textarea. it becomes noticeable when you have an error showing.

packages/FormGroup/src/FormGroup.stories.tsx
vicky-comeau: good catch, this should be fixe dint he next build

packages/FormGroup/src/FormGroup.stories.tsx
vicky-comeau: I think this should be fixed now, let me know if I missed something.

packages/FormGroup/src/FormGroup.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  const classes = cx('ids-form-group', className);
+
+  return (
+    <div className={classes} data-test={dataTest}>
+      {label && (
+        <label htmlFor={htmlFor} className="ids-form-group__label">
+          {label}
+        </label>
+      )}
```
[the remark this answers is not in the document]
franckgaudin: it's a good practice but the code is relatively simple and in my opinion doesn't need to create a method for each element. If necessary, we can simply create a `const` with the label and error message.

### Example:
```
const renderLabel = <label htmlFor={htmlFor} className="ids-form-group__label">{label}</label>
```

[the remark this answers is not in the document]
vicky-comeau: ok I'll keep these comments in mind for next time, but I think I will keep as is. Since like Frank said, it's pretty simple either way.

franckgaudin (approved):

PR #222, opened 2022-09-28
Feat: Initial release of the filter component

vicky-comeau:


packages/Filter/src/Filter.tsx
```diff
@@ -0,0 +1,54 @@
+import * as React from 'react';
+
+import cx from 'classnames';
+import './filter.scss';
+
+export interface FilterProps extends React.ComponentProps<'button'> {
+  /** Add the main text/icon to the filter */
+  children: React.ReactNode;
+  /** Add class names to the filter button */
```
ofrogon, line 9: Elsewhere we just call it `filter` (vs `filter button`) for consistency I think the `button` should be remove from the sentence.

packages/Filter/src/filter.scss
```diff
@@ -0,0 +1,106 @@
...
+  --ids-filter-color-selected: #{tokens.$samoyed};
+
+  /* Disabled */
+  --ids-filter-color-disabled: #{tokens.$grey-400};
+
+  /* Focus */
+
+  /* Needed to add a 0.1px blur to fix a bug where
+  chrome would have a lingering blue line when focusing in and out */
+  --ids-filter-focus: 0 0 0 2px #fff, 0 0 0.1px 3.5px #0c46d0;
```
ofrogon, line 28: We had fix that same issue in the `Button` maybe you can use the same solution here.

By memory it was something with also using:
```css
outline: 0;
```
in the CSS properties.

vicky-comeau: sadly this isn't the same thing. It's specifically due to the box-shadow. It's like some sort of glitch on chrome.
This is what it looks like if you're tabbing on the filter:
image

ofrogon (dismissed):

packages/Filter/src/Filter.tsx
```diff
@@ -0,0 +1,54 @@
...
+import cx from 'classnames';
+import './filter.scss';
+
+export interface FilterProps extends React.ComponentProps<'button'> {
+  /** Add the main text/icon to the filter */
+  children: React.ReactNode;
+  /** Add class names to the filter */
+  className?: string;
+  /** Count of items that match the filter */
+  count?: number;
```
franckgaudin, line 12: why not just pass the **count** via the `children` by the user?

`<Filter>Assigned to me (1)</Filter>`

franckgaudin:
I would reduce the width of the icon to match the tag in Figma and you can remove the point on the right 

filter-icons

packages/Filter/src/Filter.tsx
vicky-comeau: It's a possibility. If the count is updated, will it re-render just the count and not the whole child element?

franckgaudin: if the number changes the `Filter` component will be re-rendered

franckgaudin (approved):

ofrogon (approved):

PR #224, opened 2022-09-30
Feat: breadcrumb component release

vicky-comeau:


packages/Breadcrumb/src/Breadcrumb.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+import './breadcrumb.scss';
+
+type Item = {
+  /** Add text for the breadcrumb */
+  label: React.ReactNode;
+  link?: string;
+};
+
+export interface BreadcrumbProps extends React.ComponentProps<'ul'> {
```
franckgaudin, line 16: small detail but we use in the return a `ol`

vicky-comeau: oh shoot good catch!. I was originally going to use a ul and forgot to update it :/

franckgaudin (approved):

PR #230, opened 2022-10-14
Feat: added tooltip animation and changed default positioning to top

vicky-comeau:


.changeset/silent-birds-wonder.md
```diff
@@ -0,0 +1,5 @@
+---
+'@igloo-ui/tooltip': minor
```
franckgaudin, line 2: Changing the default position implies a major versioN. A description must also be added to the BREAKING CHANGE

franckgaudin (approved):

PR #240, opened 2022-11-07
feat(toast): Rework code and update UI

franckgaudin:


packages/Toaster/src/Toaster.test.tsx
```diff
@@ -2,51 +2,46 @@
...
-  test('It should render success', () => {
-    expect(
-      <Toast toast={{ ...mockToast, type: 'success' }} />
-    ).toMatchSnapshot();
+  test('It should render error toast', () => {
+    const { getByText } = render(
+      <Toast error message="This didn't work!" dataTest="errorToast" />
+    );
+    const toast = getByText("This didn't work!");
+    expect(toast.classList.contains('ids-toast--error')).toBe(true);
```
vicky-comeau, line 25: This is fine, but another way to do it is:
`expect(toast).toHaveClass('ids-toast--error');`

utils/generate-api.mjs
```diff
@@ -29,7 +29,7 @@ export async function getComponentAPI(source) {
     return !stats.isDirectory();
   });
 
-  const ignoreFiles = ['Toast', 'ButtonGroupItem'];
+  const ignoreFiles = ['ToasterContainer', 'Toast', 'ButtonGroupItem'];
```
vicky-comeau, line 32: More of a question for me. Why does this need to be added here?

franckgaudin: This variable allows to filter the files that should not be parsed by `react-docgen-typescript`. It avoids to have in the documentation the props relative to these files

vicky-comeau (approved):

jpineault (approved):

PR #256, opened 2022-11-16
Feat: added an ellipsis when too long, a color icon option, an image …

vicky-comeau:
…icon option, a warning icon option, and an error look


packages/Tag/src/Tag.tsx
```diff
@@ -21,44 +26,92 @@ export type Appearance =
...
   const [show, setShow] = React.useState(true);
 
-  const renderDismissButton = (
-    setShow: (show: boolean) => void,
-    onDismissClick?: () => void
-  ): JSX.Element => {
+  const renderIcon = (): JSX.Element | null => {
+    let visual = null;
+
+    if (isWarning) {
```
franckgaudin, line 78: To avoid if else if it is better to use conditional rendering.

```
    if (isWarning) {
      return (
        <IconAlert
          size="small"
          className="ids-tag__visual ids-tag__warning-icon"
        />
      );
    }

    if (icon) {
      return React.cloneElement(icon, {
        className: 'ids-tag__visual ids-tag__icon',
      });
    }

  ...

  return null
```

packages/Tag/src/Tag.tsx
```diff
@@ -21,44 +26,92 @@ export type Appearance =
...
   dismissible?: boolean;
   /** The icon used at the beginning of the Tag */
-  icon?: React.ReactNode;
+  icon?: React.ReactElement;
+  /** The id of the tag */
+  id?: string;
+  /** Add an error style to the tag */
+  hasError?: boolean;
+  /** Add a warning icon to the tag. Overrides any image, color or icon */
+  isWarning?: boolean;
```
franckgaudin, line 46: It seems that this props is never used in the portal, in my opinion you can remove it

packages/Tag/src/Tag.tsx
```diff
@@ -21,44 +26,92 @@ export type Appearance =
...
+  dataTest?: string;
   /** Render the close button */
   dismissible?: boolean;
   /** The icon used at the beginning of the Tag */
-  icon?: React.ReactNode;
+  icon?: React.ReactElement;
+  /** The id of the tag */
+  id?: string;
+  /** Add an error style to the tag */
+  hasError?: boolean;
```
franckgaudin, line 44: idem has `isWarning`

packages/Tag/src/tag.scss
```diff
@@ -49,18 +53,24 @@
 
   /* Medium */
   --ids-tag-font-size: #{tokens.$font-size-3};
-  --ids-tag-padding: calc(#{tokens.$space-4} / 4) #{tokens.$space-3};
-  --ids-tag-icon-margin: #{tokens.$space-2};
+  --ids-tag-height: 3.2rem;
```
franckgaudin, line 56: Use the token for this size `#{tokens.$space-5}`

packages/Tag/src/tag.scss
```diff
@@ -49,18 +53,24 @@
...
+  --ids-tag-padding: #{tokens.$space-0} #{tokens.$space-2};
+  --ids-tag-content-padding: #{tokens.$space-0} #{tokens.$space-2};
 
   /* Small */
   --ids-tag-font-size-small: #{tokens.$font-size-2};
-  --ids-tag-padding-small: #{tokens.$space-1} #{tokens.$space-2};
+  --ids-tag-height-small: 2.4rem;
```
franckgaudin, line 62:
```suggestion
  --ids-tag-height-small: #{tokens.$space-4};
```

packages/Tag/src/tag.scss
```diff
@@ -49,18 +53,24 @@
...
+  --ids-tag-padding-small: #{tokens.$space-0};
 
   /* XSmall */
-  --ids-tag-padding-xsmall: calc(#{tokens.$space-1} / 2) #{tokens.$space-2};
+  --ids-tag-height-xsmall: 2rem;
 
   /* Micro */
-  --ids-tag-padding-micro: #{tokens.$space-0} #{tokens.$space-2};
+  --ids-tag-height-micro: 1.6rem;
```
franckgaudin, line 69:
```suggestion
  --ids-tag-height-micro: #{tokens.$space-3};
```

packages/Tag/src/user-image.scss
```diff
@@ -0,0 +1,62 @@
...
+  --ids-user-image-font-family: #{tokens.$secondary-font-family};
+  --ids-user-image-font-size: #{tokens.$font-size-4};
+  --ids-user-image-font-weight: #{tokens.$font-weight-medium};
+
+  /* Sizes */
+  --ids-user-image-size-xs: #{tokens.$font-size-4};
+  --ids-user-image-size-s: #{tokens.$font-size-7};
+  --ids-user-image-size-m: #{tokens.$font-size-9};
+  --ids-user-image-size-l: #{tokens.$font-size-10};
+  --ids-user-image-size-xl: 6.4rem;
```
franckgaudin, line 17:
```suggestion
  --ids-user-image-size-xl: ${tokens.$space-8};
```

packages/Tag/src/TeamColorIcon.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+export type Size = 'small' | 'medium' | 'large' | 'xlarge';
+
+export interface TeamColorIconProps extends React.ComponentProps<'div'> {
+  /** The content of the team color icon */
+  children?: React.ReactNode;
+  /** Add a specific class to the team color icon */
+  className?: string;
+  /** The background color of the team icon */
+  teamColor?: string;
```
franckgaudin, line 17: why not just use `color`?

packages/Tag/src/TeamColorIcon.tsx
```diff
@@ -0,0 +1,148 @@
...
+  /** Specifies the size of the team icon */
+  size?: Size;
+  /** Specifies the id of the color tag,
+   * if the team color properties is empty,
+   * we will calculate the color with the id. */
+  teamId?: string;
```
franckgaudin, line 23: why not just use `id`?

packages/Tag/src/TeamColorIcon.tsx
```diff
@@ -0,0 +1,148 @@
...
+  size = 'small',
+  teamId,
+  teamName,
+}) => {
+  const classes = cx(
+    'ids-team-color-icon',
+    `ids-team-color-icon--${size}`,
+    className
+  );
+  const allOrganisationId = '00000000-0000-0000-0000-000000000000';
```
franckgaudin, line 42: The whole logic part that defines a color according to the id should be the responsibility of the user ex. portal. By doing so we make the component too specific to a use case. The idea here is to receive a color ex. #ccff22 and pass it to the square.

packages/Tag/src/Tag.tsx
vicky-comeau: I know this one is used in some places like here:
image

packages/Tag/src/TeamColorIcon.tsx
vicky-comeau: alright, so I'll just remove the ID prop and its logic?

franckgaudin: yes, the ID and logic

packages/Tag/src/Tag.tsx
```diff
@@ -1,8 +1,13 @@
 import * as React from 'react';
 
+import cx from 'classnames';
+
 import IconButton from '@igloo-ui/icon-button';
+import Ellipsis from '@igloo-ui/ellipsis';
```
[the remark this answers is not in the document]
vicky-comeau: now it is :)

packages/Tag/src/UserImage.tsx
```diff
@@ -0,0 +1,53 @@
...
+  return (
+    <div className={classes}>
+      <img
+        alt=""
+        className={cx(
+          'ids-user-image__image',
+          `ids-user-image__image--${size}`
+        )}
+        src={checkImgSrc()}
+        data-hj-suppress
```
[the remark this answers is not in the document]
vicky-comeau: I looked it up and it has something to do with hotjar... it was in the previous component, but I should probably just remove it, right?

ofrogon (approved):

packages/Tag/src/components/UserImage.tsx
```diff
@@ -0,0 +1,51 @@
...
+      image = `data:image/jpeg;base64,${image}`;
+    }
+
+    return image;
+  };
+
+  return (
+    <div className={classes}>
+      <img
+        alt=""
```
franckgaudin, line 40: in a next iteration it would be nice to pass a props and a default value for the `alt`

franckgaudin (approved):

PR #263, opened 2022-11-24
Chore/shared component

franckgaudin:
The shared components have been created to help the development of Igloo, they are not intended to be published for the moment. 

the `VisualIdentifier` component is here as an example, it will be reworked soon when the `Tag` component is modified


shared/components/src/Index.tsx
```diff
@@ -0,0 +1,3 @@
+import * as React from 'react';
+
+export { VisualIdentifier } from './visual-identifier/VisualIdentifier';
```
vicky-comeau, line 3: Just a question. For each shared component, will we need to add an export here?

franckgaudin: Yes! 
Shared Components should remain in special cases, we shouldn't have too many of them.

vicky-comeau (approved):

jpineault (approved):

PR #265, opened 2022-11-24
Refactor: Extracted Avatar and Color from Tag to create new components

vicky-comeau:


packages/Color/src/Color.tsx
```diff
@@ -0,0 +1,85 @@
...
+          break;
+        }
+      }
+
+      if (initials.length < 2) {
+        initials = splitName[0].slice(0, 2);
+      }
+    }
+
+    if (size === 'small' || size === 'medium') {
```
franckgaudin, line 67: I did not see this in Figma, do you have an example of use?

packages/Avatar/src/Avatar.tsx
```diff
@@ -0,0 +1,57 @@
+import * as React from 'react';
+
+import cx from 'classnames';
+
+import './avatar.scss';
+
+export type Size = 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge';
```
franckgaudin, line 7: I couldn't find a reference to the 'xlarge' size, do you have an example of its use?

packages/Color/src/Color.tsx
```diff
@@ -0,0 +1,85 @@
+import * as React from 'react';
+
+import cx from 'classnames';
+
+import './color.scss';
+
+export type Size = 'small' | 'medium' | 'large' | 'xlarge';
```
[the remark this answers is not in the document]
vicky-comeau: Not sure. I just followed the previous code for the size. I could ask Francis

packages/Avatar/src/Avatar.stories.tsx
```diff
@@ -0,0 +1,55 @@
...
+export const Overview = Template.bind({});
+Overview.args = {
+  src: 'https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcQEB3tCC4oJsa4ZZsiEDIhWi06EyN3iHYMoOg&usqp=CAU',
+};
+
+export const Sizes = () => (
+  <Section>
+    <Avatar
+      size="xsmall"
+      src="https://cdn-icons-png.flaticon.com/512/168/168724.png"
```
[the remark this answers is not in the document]
vicky-comeau: The only reason I didn't use those ones it because the corners already seemed rounded. I wanted something not rounded so my border-radius would show.

packages/Color/src/Color.tsx
vicky-comeau: This was also in the previous code so I opted to keep it.

packages/Avatar/src/Avatar.tsx
vicky-comeau: In the old code, it seems to be used in TeamMemberHeader
image

packages/Color/src/Color.tsx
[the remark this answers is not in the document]
franckgaudin: Good idea, you could also see with him to add the size 64px `xlarge` of the avatar.

ofrogon (approved):

franckgaudin (approved):

PR #271, opened 2022-11-29
feat(BarChart): add animation

franckgaudin:


packages/BarChart/src/__snapshots__/BarChart.test.tsx.snap
```diff
@@ -20,7 +20,7 @@ exports[`BarChart It should render a snapshot 1`] = `
         <div
           class="ids-bar-chart__graph"
           data-value="48"
-          style="width: 100%; background: rgb(251, 199, 59);"
+          style="width: 0px; background-color: rgb(251, 199, 59);"
```
vicky-comeau, line 23: px shouldn't be needed when 0. I'd suggest removing it.

packages/BarChart/src/__snapshots__/BarChart.test.tsx.snap
```diff
@@ -43,7 +43,7 @@ exports[`BarChart It should render a snapshot 1`] = `
         <div
           class="ids-bar-chart__graph"
           data-value="46"
-          style="width: 95.83333333333333%; background: rgb(16, 83, 255);"
+          style="width: 0px; background-color: rgb(16, 83, 255);"
```
vicky-comeau, line 46: same as above

packages/BarChart/src/__snapshots__/BarChart.test.tsx.snap
```diff
@@ -66,7 +66,7 @@ exports[`BarChart It should render a snapshot 1`] = `
         <div
           class="ids-bar-chart__graph"
           data-value="24"
-          style="width: 50%; background: rgb(231, 145, 154);"
+          style="width: 0px; background-color: rgb(231, 145, 154);"
```
vicky-comeau, line 69: here too

vicky-comeau:
I also noticed that if you change any props in the dataSet, it will reload the whole chart and the animation will run again. Is there a way that the animation will only run what the value changes? To go even deeper, it could run only for the specific bar that the value was changed and not always from 0. If the value was 50 and you set it to 75, then it should animate from 50 to 75.

packages/BarChart/src/__snapshots__/BarChart.test.tsx.snap
franckgaudin: the value is setter by `react-spring`

packages/BarChart/src/bar-chart.scss
```diff
@@ -52,12 +53,8 @@
 
   &__graph {
     height: var(--ids-bar-chart-graph-height);
-    width: var(--ids-bar-chart-graph-width, 100%);
+    width: var(--ids-bar-chart-graph-width, var(--ids-bar-chart-graph-width));
     border-radius: var(--ids-bar-chart-graph-radius);
     background: var(--ids-bar-chart-graph-background);
-
-    &[data-value='0'] {
-      --ids-bar-chart-graph-width: #{tokens.$space-1};
```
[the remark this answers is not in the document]
franckgaudin: errors in the dev 🤯

packages/BarChart/src/__snapshots__/BarChart.test.tsx.snap
vicky-comeau: oh right, it's the snapshot, sorry

franckgaudin:
> I also noticed that if you change any props in the dataSet, it will reload the whole chart and the animation will run again. Is there a way that the animation will only run what the value changes? To go even deeper, it could run only for the specific bar that the value was changed and not always from 0. If the value was 50 and you set it to 75, then it should animate from 50 to 75.


https://user-images.githubusercontent.com/2379339/205365637-08ce452d-1be0-4dab-8a2b-3c8362c3b5b0.mov

packages/Tag/src/Tag.test.tsx
```diff
@@ -132,7 +132,6 @@ describe('Tag', () => {
       src: 'https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcQEB3tCC4oJsa4ZZsiEDIhWi06EyN3iHYMoOg&usqp=CAU',
     });
     const icon = container.querySelector('.ids-tag__visual');
-    console.log(icon);
```
vicky-comeau, line 135: shoot was that something I left? ;/

franckgaudin: it's by chance that I saw this

vicky-comeau:
> > I also noticed that if you change any props in the dataSet, it will reload the whole chart and the animation will run again. Is there a way that the animation will only run what the value changes? To go even deeper, it could run only for the specific bar that the value was changed and not always from 0. If the value was 50 and you set it to 75, then it should animate from 50 to 75.
> 
>  barChart-animation.mov

Looks good. I'm guessing you had to edit it with your own component because storybook just always re-renders when a property is changed?

packages/BarChart/src/BarChart.tsx
```diff
@@ -26,9 +30,13 @@ const BarChart: React.FunctionComponent<BarChartProps> = (
...
-  if (!dataSet) {
-    return null;
-  }
+  const [animation, setAnimation] = React.useState(false);
+
+  React.useEffect(() => {
+    setAnimation(!animation);
+    const timer = setTimeout(() => setAnimation(false), 300);
+    return () => clearTimeout(timer);
+  }, []);
```
ofrogon, line 39: `animation` should be a dependency.

ofrogon:
```suggestion
  }, [animation]);
```

franckgaudin: Adding `animation` in dependency will create an infinite loop, the case of figure which I wish here it is to change that value at the time when the component is mounted only.

franckgaudin:
> Looks good. I'm guessing you had to edit it with your own component because storybook just always re-renders when a property is changed?

in fact it was easier for me to test it this way 😄

packages/BarChart/src/BarChart.tsx
vicky-comeau: ya the warnings we have for the dependency arrays are annoying :( sometimes you just want it to be empty to only run on mount! but it doesn't like that :(

ofrogon (approved):

PR #336, opened 2023-02-16
Feat: [OV-41453] Initial release of disclosure

vicky-comeau:


packages/Disclosure/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "rollup -c rollup.config.js"
+  },
+  "dependencies": {
+    "@igloo-ui/tokens": "^2.0.0",
+    "classnames": "^2.3.2",
+    "framer-motion": "^6.5.1",
```
code-person, line 29: Is there a reason why we're using framer-motion instead of react-spring here?

vicky-comeau: Yes, framer motion had an easy built in way to animate to auto height. I was not able to do this with react-spring. I've been running into a lot of little bugs with react-spring as well so I tried something different.

ofrogon (dismissed):

packages/Disclosure/src/disclosure.scss
```diff
@@ -0,0 +1,107 @@
...
+.ids-disclosure {
+  box-shadow: var(--ids-disclosure-shadow);
+  font-family: var(--ids-disclosure-font-family);
+  font-size: var(--ids-disclosure-font-size);
+
+  .ids-disclosure__header {
+    align-items: center;
+    background: var(--ids-disclosure-header-background);
+    border: none;
+    border-radius: var(--ids-disclosure-border-radius);
```
franckgaudin, line 40: once opened there should be no border radius at the bottom of the button.

Capture d’écran, le 2023-02-21 à 16 46 29

franckgaudin (commented):
when the component is open the arrow must point up. 

**Docs**
Capture d’écran, le 2023-02-21 à 12 38 16

**Figma**
Capture d’écran, le 2023-02-21 à 12 41 41

packages/Disclosure/src/disclosure.scss
vicky-comeau: ya that was supposed to be on the ids-disclosure element. Good catch!

ofrogon (approved):

PR #353, opened 2023-03-03
chore: Upgrade project for using Node v18

franckgaudin:
- config storybook for using webpack 5
- upgrade storybook to the last version.

ticket OV-42041


package.json
```diff
@@ -21,7 +21,7 @@
     "lint:all": "yarn lint && yarn lint:style",
     "fix:all": "yarn lint:fix && yarn fix:style",
     "doc:start": "lerna run dev --scope=docs",
-    "doc:build": "lerna run build --verbose --scope=docs && NODE_OPTIONS=--openssl-legacy-provider yarn build-storybook --docs -o ./website/.next/storybook",
```
vicky-comeau, line 24: This isn't needed anymore? so would I remove cross-env?

vicky-comeau (approved):

package.json
franckgaudin: indeed by switching to webpack 5 it is no longer necessary

ofrogon (approved):

PR #371, opened 2023-03-22
docs: Adding a troubleshooting section in the doc

franckgaudin:
We explain how to work around the problem of the output on the modal and the dialog component

OV-42443

### DIFF
--- a/packages/Dialog/README.md
+++ b/packages/Dialog/README.md
@@ -47,3 +47,11 @@ const handleValidate = () => {
     onValidate={handleValidate}
 />
 ```
+
+## Troubleshooting
+
+**When a Dialog opens above a Modal**
+
+In this case when you close the Dialog the Modal closes automatically, here is a workaround to remedy this.
+
+To solve the problem set the `isClosable` property of the Modal to `false` when the Dialog component is up and then return to `true` when it closes.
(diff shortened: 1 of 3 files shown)


packages/Dialog/README.md
```diff
@@ -47,3 +47,11 @@ const handleValidate = () => {
...
 />
 ```
+
+## Troubleshooting
+
+**When a Dialog opens above a Modal**
+
+In this case when you close the Dialog the Modal closes automatically, here is a workaround to remedy this.
+
+To solve the problem set the `isClosable` property of the Modal to `false` when the Dialog component is up and then return to `true` when it closes.
```
vicky-comeau, line 57:
```suggestion
To solve the problem, set the `isClosable` property of the Modal to `false` when the Dialog component is up and then return to `true` when it closes.
```

packages/Modal/README.md
```diff
@@ -45,3 +45,11 @@ const [show, setShow] = useState(false);
...
 </Modal>;
 ```
+
+## Troubleshooting
+
+**When a Dialog opens above a Modal**
+
+In this case when you close the Dialog the Modal closes automatically, here is a workaround to remedy this.
+
+To solve the problem set the `isClosable` property of the Modal to `false` when the Dialog component is up and then return to `true` when it closes.
```
vicky-comeau, line 55:
```suggestion
To solve the problem, set the `isClosable` property of the Modal to `false` when the Dialog component is up and then return to `true` when it closes.
```

vicky-comeau (approved):

ofrogon (approved):

PR #380, opened 2023-03-30
Fix: [OV-42779] Modal fixes when using keys

vicky-comeau:


packages/Modal/src/Modal.tsx
```diff
@@ -83,6 +85,7 @@ const Modal: React.FunctionComponent<ModalProps> = (props: ModalProps) => {
     primaryAction,
     secondaryAction,
     carousel,
+    keyValue = '',
```
franckgaudin, line 88: Why not use an ID here

franckgaudin (approved):

ofrogon (approved):

PR #386, opened 2023-04-12
feat: add new options for disabled date and new callback for clear date

franckgaudin:
OV-42656


ofrogon (approved):

vicky-comeau:
We are not seeing the right controls for the new props. The first should be a boolean and not a date and the minValue should be a date field I imagine: 
image

vicky-comeau (approved):

ofrogon (approved):

Issue #406, opened 2023-05-25
[Bug]: Datepicker with unavailable weekend cursor problem


CarolineSabourin-Jean:
### What happened?

The cursor is flickering when hovering the disabled weekends:
![datepicker](https://github.com/gsoft-inc/ov-igloo-ui/assets/25105785/a7952a7c-b54b-40e9-a81e-72c3bb576690)

### Component

Datepicker

### Component Version

0.1.1

### Which browsers are you seeing the problem on?

Chrome

PR #407, opened 2023-05-25
fix: cursor display on hover of an inactive date

franckgaudin:
OV-43773 
issue: #406

### DIFF
--- a/packages/Datepicker/src/datepicker.scss
+++ b/packages/Datepicker/src/datepicker.scss
@@ -144,6 +144,11 @@ $datepicker-width: calc($cell-number * $cell-size - $cell-border - $cell-empty-b
 
     &.ids-datepicker__content--selected {
       color: var(--ids-datepicker-content-color-selected);
+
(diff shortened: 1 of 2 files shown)


packages/Datepicker/src/datepicker.scss
```diff
@@ -117,8 +122,8 @@ $datepicker-width: calc($cell-number * $cell-size - $cell-border - $cell-empty-b
   margin-top: -1px;
 
   &:hover {
-    background: var(--ids-datepicker-content-background-hover);
-    color: var(--ids-datepicker-content-color-hover);
+    @extend %active-date;
```
vicky-comeau, line 125: I see you switched the hover colors for the selected one. Is this the desired effect? If so, please remove the variable definitions above.

franckgaudin: Good point 👍🏻

vicky-comeau (approved):

PR #409, opened 2023-05-26
feat: [OV-43587] Added char limit

vicky-comeau:


packages/Input/src/input.scss
```diff
@@ -33,6 +33,15 @@
   --ids-input-width-suffix: #{tokens.$space-7};
   --ids-input-width-suffix-compact: #{tokens.$space-6};
   --ids-input-padding-suffix: #{tokens.$space-2} #{tokens.$space-3} #{tokens.$space-2} #{tokens.$space-2};
+  --ids-input-padding-suffix-with-indicator: #{tokens.$space-2} #{tokens.$space-2} #{tokens.$space-2} #{tokens.$space-2};
```
ofrogon, line 36: `--ids-input-padding-suffix-with-indicator` is unused.

packages/Input/src/Input.tsx
```diff
@@ -143,11 +168,25 @@ const Input: React.FunctionComponent<InputProps> = React.forwardRef(
...
+          <div
+            className={cx('ids-input__asset-wrapper', {
+              'ids-input__asset-wrapper--has-char-indicator':
+                displayCharIndicator,
+            })}
+          >
+            {suffixIcon && addSuffixIcon}
+            {displayCharIndicator && (
+              <div className="ids-input__char-indicator">
+                {inputMaxLength - charLength}
```
ofrogon, line 185: Their is something with the character count that act strange. For some reason the length of `😀😀` is the same as `😀😀a`

Visually:
image
image

vicky-comeau: It seems like the original fix we had for emojis is no longer needed. It was now causing issues.

franckgaudin:
The issue with emojis is unfortunately still with us. 
If we add an emoji when there is 1 character left, the remaining number will be -1.

input-emoji

franckgaudin (approved):

ofrogon (approved):

PR #415, opened 2023-06-02
Feat: [OV-43759] TagPicker added search icon and error prop

vicky-comeau:


ofrogon (approved):

franckgaudin:
There seems to be a problem with the input following the changes.

https://github.com/gsoft-inc/ov-igloo-ui/assets/2379339/3da50327-6c2f-44d3-8fd3-a5f5fbaf76bc

ofrogon:
Hi, while testing the `TagPicker` I encounter a strange result.

Filter request seems to occurred even if the dropdown is disappearing.

To reproduce:
- In one of the `TagPicker` example with the large Team list.
- Enter `Te` (to have the filtered list displayed).
- Delete the last character.
  - Thile the list is going away, the result `This user's name is very looooooong` will show during transition.

How I think this should work is that result in the dropdown should now update if the dropdown is fading away.

image
image

franckgaudin:
Is this the UI we want if a tag is already present in the input?

tag-picker-search

PR #422, opened 2023-06-07
Center modal using flex instructions

ThiBsc:
### Contribution reason
The CSS `transform` instruction used to center the modal make a position offset when using drag&drop with `react-beautiful-dnd` inside a modal.

### Changes
- Wrapping the modal in a div with `className="ids-modal-wrapper"`
- Remove transform and fixed positioning on the modal
- Center the modal inside the wrapper by using flex instruction in the wrapper

### Tests
```diff
+ Test Suites: 41 passed, 41 total
+ Tests:       257 passed, 257 total
+ Snapshots:   59 passed, 59 total
# Time:        18.473 s
# Ran all test suites.
```

### DIFF
--- a/packages/Modal/src/Modal.tsx
+++ b/packages/Modal/src/Modal.tsx
@@ -160,7 +160,7 @@ const Modal: React.FunctionComponent<ModalProps> = (props: ModalProps) => {
       </AnimatePresence>
       <AnimatePresence onExitComplete={onExitComplete}>
         {isOpen && (
-          <m.div className="ids-modal-wrapper">
+          <m.div className="ids-modal__wrapper">
             <m.div
               key={`${keyValue}_modal`}
               className={classes}
--- a/packages/Modal/src/Modal.tsx
+++ b/packages/Modal/src/Modal.tsx
@@ -121,20 +121,14 @@ const Modal: React.FunctionComponent<ModalProps> = (props: ModalProps) => {
     open: {
       opacity: 1,
       scale: 1,
-      x: '-50%',
-      y: '-50%',
     },
     close: {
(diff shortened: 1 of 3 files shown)


packages/Modal/src/Modal.tsx
```diff
@@ -166,88 +160,90 @@ const Modal: React.FunctionComponent<ModalProps> = (props: ModalProps) => {
...
-            exit="close"
-            variants={modalVariants}
-            transition={{ duration: 0.2 }}
-            ref={ref}
-          >
-            <div
-              className={cx('ids-modal__header', {
-                'ids-modal__header--with-back-btn': displayBackBtn,
-              })}
+          <m.div className="ids-modal-wrapper">
```
vicky-comeau, line 163: I think it should follow the way we name the others like ids-modal__header so it should be:
```suggestion
          <m.div className="ids-modal__wrapper">
```

packages/Modal/src/modal.scss
```diff
@@ -78,6 +78,18 @@
   z-index: var(--ids-modal-overlay-index);
 }
 
+.ids-modal-wrapper {
```
vicky-comeau, line 81:
```suggestion
.ids-modal__wrapper {
```

packages/Modal/src/Modal.tsx
ThiBsc: I'm not sure in that case, because `__something` is more when an element is child of, but in this case, wrapper is not the child of the modal

vicky-comeau (approved):

franckgaudin (approved):

PR #433, opened 2023-06-12
Feat: [OV-43388] Added icon prop

vicky-comeau:


ofrogon (dismissed):

franckgaudin:
Is the size of the custom icon managed in the case of a custom button?

vicky-comeau:
> Is the size of the custom icon managed in the case of a custom button?

Good point, I didn't try and see what happens if you put different sized icons.
Should the user have pull control on the size? or do you think we should manage it?

ofrogon (approved):

franckgaudin (approved):

PR #436, opened 2023-06-14
Feat: [OV-43613] Added new props to combobox and list

vicky-comeau:


ofrogon (dismissed):

packages/Combobox/src/Combobox.stories.tsx
```diff
@@ -276,6 +297,48 @@ export const Multiple: Story = {
...
+      </Section>
+    )
+  },
+
+  parameters: {
+    chromatic: { disableSnapshot: true },
+  },
+};
+
+export const WithFooter: Story = {
```
franckgaudin, line 321: I don't see the story in the preview, is that normal?

franckgaudin (commented):
The button is not displayed when navigating the `combobox` with the keyboard. 
It is also not possible to access the button in the footer. 

combobox-focus

vicky-comeau:
> The button is not displayed when navigating the `combobox` with the keyboard. It is also not possible to access the button in the footer.
> 
> combobox-focus

I'm not sure what to do in this case. I could allow tabbing and it'll work, but it's a little odd that for the options you use arrow keys and the links you have to use tabbing. Do you have any other ideas? Just to make sure, does it work for you in the List component?

franckgaudin:
I'm not sure how to treat it - you'd have to do research for that. The list component works with keyboard navigation.

ofrogon (approved):

franckgaudin (approved):

PR #439, opened 2023-06-16
chore(doc): Fix the size of table of content

franckgaudin:
#OV-44186


vicky-comeau:
I know it's not part of the original issue, but it's related. When resizing the screen just before the left menu disappears, you will see some overlap:
image

ofrogon (commented):
PR code is great! But you may want to fix what @vicky-comeau flag at the same time.

ofrogon (approved):

vicky-comeau (approved):

PR #442, opened 2023-06-19
Add props for reset the size to the default on mobile

franckgaudin:
Ticket #OV-44028

### DIFF
--- a/packages/Button/src/Button.tsx
+++ b/packages/Button/src/Button.tsx
@@ -64,6 +64,8 @@ export interface ButtonOwnProps {
   intercomTarget?: string;
   /** Add a specific class to the button */
   className?: string;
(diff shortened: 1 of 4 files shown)


vicky-comeau (dismissed):

packages/Button/src/Button.tsx
```diff
@@ -64,6 +64,8 @@ export interface ButtonOwnProps {
   intercomTarget?: string;
   /** Add a specific class to the button */
   className?: string;
+  //* * Return to the default size on mobile */
```
ofrogon, line 67:
```suggestion
  /** Return to the default size on mobile */
```

ofrogon (dismissed):

ofrogon (approved):

PR #463, opened 2023-07-18
Feat: [OV-45016] Pie Chart Initial Release

vicky-comeau:


packages/PieChart/src/PieChart.tsx
```diff
@@ -0,0 +1,135 @@
...
+      </tspan>
+      <tspan
+        baselineShift="super"
+        alignmentBaseline="middle"
+        className="ids-pie-chart__symbol"
+      >
+        %
+      </tspan>
+      <tspan x={cx} dy={halfOfRateHeight} className="ids-pie-chart__label">
+        {label}
```
ofrogon, line 70: Do we want to limit the size of the label on support ellipsis for situations like this:
image

packages/PieChart/src/pie-chart.scss
```diff
@@ -0,0 +1,48 @@
...
+}
+
+.ids-pie-chart__symbol {
+  fill: var(--ids-pie-chart-symbol-color);
+  font-size: var(--ids-pie-chart-symbol-font-size);
+}
+
+.ids-pie-chart__label {
+  fill: var(--ids-pie-chart-label-color);
+
```
ofrogon, line 46: You can remove this empty line.

franckgaudin:
Small display problem in the doc. 

Capture d’écran, le 2023-07-18 à 15 05 12

packages/PieChart/src/PieChart.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+     const {asFragment} = setup();
+     expect(asFragment()).toMatchSnapshot();
+   });
+
+   test('It should render the provided data elements', () => {
+    const data = [
+      { id: 'data1', percentage: 30 },
+      { id: 'data2', percentage: 20 },
+      { id: 'data3', percentage: 50 },
+    ];
```
ofrogon, line 32: Coud be nice to also test when the `sum > 100` and `sum < 100`

vicky-comeau:
> Small display problem in the doc.
> 
> Capture d’écran, le 2023-07-18 à 15 05 12

Must be a Mac bug, I'll debug it, thanks.

packages/PieChart/src/PieChart.tsx
vicky-comeau: Good point. For percentage, we could probably limit the number between 0-100.
Should the label actually have a character limit, or just put an ellipsis using css? I vote for css, since the label can be anything, not just a string. What do you think?

ofrogon: I vote for CSS too.

packages/PieChart/src/PieChart.test.tsx
vicky-comeau: I'm not sure how this benefits us in the tests? The data will be the same, it's not like we'll be testing the user data.
Also, recharts will always fill the chart. For example, if it adds up to only 60, it thinks the total is 60 and displays it accordingly. I guess it's up to the user to put the proper values. Unless you have another suggestion?

packages/PieChart/src/PieChart.tsx
vicky-comeau: well I couldn't find a way to actually type the percentage so the user will just need to type it in properly. They shouldn't go above 100

packages/PieChart/src/PieChart.test.tsx
ofrogon: Thinking back on it, I was thinking more of integration tests then unit tests, where we will validate that `recharts` will always fill the chart.

Overall my comment is not really relevant since we don't test integrations.

ofrogon (approved):

franckgaudin (approved):

PR #489, opened 2023-08-08
build: [OV-45752] Storybook upgrade

vicky-comeau:


packages/List/src/list-item.scss
```diff
@@ -48,8 +48,8 @@
   --ids-list-item-checkbox-border: #{tokens.$grey-400};
   --ids-list-item-checkbox-border-hover: #{tokens.$grey-800};
   --ids-list-item-checkbox-border-checked: #{tokens.$electric-blue-500};
-  --ids-list-item-checkbox-check-width: 4px;
-  --ids-list-item-checkbox-check-height: 8px;
+  --ids-list-item-checkbox-check-width: .4rem;
+  --ids-list-item-checkbox-check-height: .4rem;
```
ofrogon, line 52: Is it normal this value changed in the process?

packages/Toaster/src/Toaster.test.tsx
```diff
@@ -154,19 +154,19 @@ describe('Toast', () => {
...
-    act(() => userEvent.unhover(toast));
+    await userEvent.unhover(toast);
 
     act(() => jest.advanceTimersByTime(4000));
     expect(toast).toHaveAttribute('data-animation', 'exiting');
 
     fireAnimationEnd(toast);
     expect(queryByRole('alert')).toBeNull();
-  });
+  }); */
```
ofrogon, line 169: Not a fan of publishing commented code... can we delete it?

vicky-comeau: It's just because I couldn't get this test to work properly with the new updates. I commented it out so that we could possibly make it work in the future. I suppose we could just remove it though. We probably want to go over and add more tests in the future anyway so it can be looked at during this time.

packages/List/src/list-item.scss
vicky-comeau: shoot, no good catch! it should be .8rem

ofrogon (approved):

franckgaudin (approved):

PR #499, opened 2023-08-22
Fix: [OV-46194] [OV-46200] [OV-46193] [OV-46190] Datepicker fixes

vicky-comeau:


packages/Datepicker/src/Datepicker.stories.tsx
```diff
@@ -131,6 +130,25 @@ export const Disabled: Story = {
...
+  },
+};
+
+export const French: Story = {
+  render: Template,
+
+  args: {
+    placeholder: 'Select date',
+    locale: 'fr-CA',
+    weekendUnavailable: true
```
fraincs, line 148: Wondering what does weekendUnavailable has to do with the French story? This seems to have two concerns bundled into one.

packages/Datepicker/src/Datepicker.tsx
```diff
@@ -1,193 +1,220 @@
...
-  minDate?: string;
-  /** The maximum allowed date that a user may select */
-  maxDate?: string;
+    selectedDay?: string;
+    /** Specifies the value inside the input. */
+    value?: string;
+    /** Text that appears in the form control when it has no value set. */
+    placeholder?: string;
+    /** Defines a string value that labels the current element. */
+    ariaLabel?: string;
```
fraincs, line 35: Should this one be enforced for the sake of accessibility?

packages/Datepicker/src/Datepicker.tsx
```diff
@@ -1,193 +1,220 @@
...
-      />
-    </Dropdown>
-  );
+    return (
+        <I18nProvider locale={locale}>
+            <Dropdown
+                isOpen={!disabled && isOpen}
+                onClose={onClose}
+                content={calendar}
+                size="medium"
```
fraincs, line 210: what's the default dropdown size? is it medium?

packages/Datepicker/src/components/Calendar.tsx
```diff
@@ -1,71 +1,72 @@
...
+        <div
+            {...calendarProps}
+            ref={ref}
+            className={className}
+            data-test={dataTest}
+        >
+            <div className="ids-datepicker__header">
+                <h2 className="ids-datepicker__title">{title}</h2>
+                <CalendarButton {...prevButtonProps} className="ids-datepicker__prev">
+                    <ChevronLeft size="medium" />
```
fraincs, line 61: Is the size needed?

packages/Datepicker/src/components/CalendarGrid.tsx
```diff
@@ -1,64 +1,64 @@
...
-import CalendarCell from './CalendarCell';
+import CalendarCell from "./CalendarCell";
 
 type State = CalendarState | RangeCalendarState;
 
 interface CalendarGridProps {
-  state: State;
-  highlightToday?: boolean;
+    state: State;
+    highlightToday?: boolean;
```
fraincs, line 13: I would have named this highlightCurrentDay

packages/Datepicker/src/Datepicker.stories.tsx
vicky-comeau: I added this here because in the code, it seemed to use locale to calculate the unavailable weekends so I thought it would be good to add to make sure it works properly.

packages/Datepicker/src/Datepicker.tsx
vicky-comeau: the default is xsmall, so this is still valid. Keep in mind, some of these files look like they have a lot of changes because when I save them, it updates with the new formatting rules we added recently to be like hopper.

packages/Datepicker/src/components/Calendar.tsx
vicky-comeau: yes. we get it from here which always adds a size https://igloo.officevibe.design/icons?search=%3DChevronLeft&size=medium
but I think the default size is small.

packages/Datepicker/src/components/CalendarGrid.tsx
vicky-comeau: What is your reasoning? I know that Datepickers tend to use the word "Today". This was previously named by Franck. I think I will keep as is for now. If you feel strongly about this, we can ask to rename it once he comes back.

fraincs: No strong opinion. Although today could not be today for someone elsewhere versus current day feels more "universal". Anyhow just a thought ;)

packages/Datepicker/src/components/Calendar.tsx
fraincs: Alright I am used with Orbit where medium is the default size.

packages/Datepicker/src/Datepicker.stories.tsx
fraincs: I get it, I would split those if it was for visual regression testing, it's ok in this situation.

PR #507, opened 2023-08-29
Fix: [OV-46424] Textarea update to fix autoResize

vicky-comeau:


packages/Textarea/package.json
```diff
@@ -26,7 +26,7 @@
   "dependencies": {
     "@igloo-ui/tokens": "^2.0.0",
     "@react-aria/utils": "^3.15.0",
-    "autosize": "^5.0.1",
+    "autosize": "^6.0.1",
```
fraincs, line 29: 6.0 dropped the support for IE. Aren't you guys still supporting IE?

Was the issue  auto resizing not happening with preloaded text solution was to update do autosize 6.0?

vicky-comeau: I don't think we are supporting IE anymore. I remember asking about this before and Franck gave me the go ahead. Once I updated the version of this, everything seemed to work fine.

fraincs (approved):

ofrogon (approved):

PR #515, opened 2023-09-06
Feature/textarea rebrand

fraincs:
- Applied the Hopper Brand to the Textarea component.


packages/Textarea/src/textarea.scss
```diff
@@ -2,94 +2,145 @@
...
+        bottom: var(--ids-textarea-char-indicator-bottom);
+        color: var(--ids-textarea-char-indicator-color);
+        font-size: var(--ids-textarea-char-indicator-size);
+        line-height: var(--ids-textarea-char-indicator-line-height);
+        position: absolute;
+        right: var(--ids-textarea-char-indicator-right);
     }
+}
 
-    &--has-char-count {
```
vicky-comeau, line 82: why was this removed?

packages/Textarea/src/textarea.scss
```diff
@@ -2,94 +2,145 @@
...
-    color: var(--ids-textarea-char-indicator-color);
-    font-size: var(--ids-textarea-char-indicator-size);
-    line-height: var(--ids-textarea-char-indicator-line-height);
-    position: absolute;
-    right: var(--ids-textarea-char-indicator-right);
-  }
+[data-brand="workleap"] .ids-textarea__field:focus,
+[data-brand="workleap"] .ids-textarea--focus .ids-textarea__field {
+    box-shadow: var(--ids-textarea-focus-box-shadow);
 }
```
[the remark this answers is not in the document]
vicky-comeau: actually, igloo should have a focus style. I'm confused as to why you only specify workleap here

vicky-comeau:
I don't think this is normal... 
image
It's on the docs so the file is Textarea.demo.js. You can start the docs locally by running 'yarn doc:start'

packages/Textarea/src/textarea.scss
vicky-comeau: It was there to prevent this overlap:
image

packages/Textarea/src/textarea.scss
```diff
@@ -2,94 +2,145 @@
...
+        &::placeholder {
+            color: var(--ids-textarea-color-placeholder);
+        }
+
+        &:focus:not(&--error),
+        .ids-textarea--focus:not(&--error) & {
+            border-color: var(--ids-textarea-focus-border-color);
+        }
+
+        &:hover:not(:disabled) {
```
vicky-comeau, line 113: I prefer not to use :not if it's not needed and can just be overridden normally. The reason being that it becomes more specific and can end up being more annoying later down the road when others use it.

franckgaudin:
> I don't think this is normal...
> image
> It's on the docs so the file is Textarea.demo.js. You can start the docs locally by running 'yarn doc:start'
> It's also visible in storybook for Igloo.

It's not just in the doc, storybook also has this error 

textarea-rebrand

packages/Textarea/src/textarea.scss
```diff
@@ -2,94 +2,145 @@
...
+    /* Error */
+    --ids-textarea-error-border-color: var(--hop-danger-border-strong);
+
+    /* Disabled */
+    --ids-textarea-disabled-background: var(--hop-neutral-border-disabled);
+    --ids-textarea-disabled-color: var(--hop-neutral-surface-disabled);
+
+    /* Character Indicator */
+    --ids-textarea-char-indicator-bottom: 0.5rem;
+    --ids-textarea-char-indicator-right: 1rem;
```
vicky-comeau, line 73: bottom and right are similar to margin. I would use a token for this.

packages/Textarea/src/textarea.scss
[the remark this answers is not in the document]
vicky-comeau: ok igloo uses the border, but my first comment still stands :P

packages/Textarea/src/textarea.scss
fraincs: I agree if these emulate a margin I am ok with it.

packages/Textarea/src/textarea.scss
fraincs: I get where you are coming from. Not is my best friend especially for disabled state, I'll override the disabled hover, I tend to prefer to avoid adding unnecessary CSS but I get the why.

packages/Textarea/src/textarea.scss
```diff
@@ -2,94 +2,156 @@
...
+        font-size: var(--ids-textarea-char-indicator-size);
+        line-height: var(--ids-textarea-char-indicator-line-height);
+        position: absolute;
+        right: var(--ids-textarea-char-indicator-right);
     }
-  }
-
-  &__char-indicator {
-    bottom: var(--ids-textarea-char-indicator-bottom);
-    color: var(--ids-textarea-char-indicator-color);
```
vicky-comeau, line 89: Do we not need a color anymore? If this is not needed, please remove the variable at the top.

fraincs: good catch I removed it by mistake.

vicky-comeau (dismissed):

vicky-comeau (approved):

franckgaudin (approved):

PR #520, opened 2023-09-07
Feat: [OV-46007] Card rebrand

vicky-comeau:


packages/Card/src/card.scss
```diff
@@ -20,13 +25,35 @@
...
+  --ids-card-padding-small: var(--hop-space-inset-md);
+
+  /* xSmall */
+  --ids-card-padding-xsmall: 0;
+}
+
 .ids-card {
   font-family: var(--ids-card-font-family);
-  padding: var(--ids-card-padding);
+  padding: var(--ids-card-padding-mobile);
```
fraincs, line 51: That's not bad here as there's only one but looking at modal/dialog moving all tokens at root and adding -mobile would create a lot of variables. Is that what we want?

fraincs (approved):

packages/Card/src/card.scss
vicky-comeau: well I feel like all workleap tokens should be at the top so we can easily find them. It was either this or I call the other one padding-large and then keep the reset of --ids-card-padding in the media query and reset it to the large padding.

fraincs: Yes I understand. The in between state has some downsides.

franckgaudin (approved):

PR #527, opened 2023-09-13
Feature/pager

fraincs:
- Styled Pager according to new brand design.

Ticket: IGL-74

### DIFF
--- a/packages/Pager/src/pager.scss
+++ b/packages/Pager/src/pager.scss
@@ -1,192 +1,287 @@
 @use '~@igloo-ui/tokens/dist/base10/variables' as tokens;
 @use '~@igloo-ui/tokens/dist/fonts';
 
+@import url('~@hopper-ui/tokens/fonts.css');
+@import url('~@hopper-ui/tokens/tokens.css');
+
 :root {
-  /* Default */
-  --ids-pager-font-family: #{tokens.$primary-font-family};
-  --ids-pager-font-size: #{tokens.$font-size-4};
-  --ids-pager-height: #{tokens.$space-5};
-  --ids-pager-line-height: #{tokens.$line-height-xxxl};
+    /* Default */
+    --ids-pager-font-family: #{tokens.$primary-font-family};
+    --ids-pager-font-size: #{tokens.$font-size-4};
+    --ids-pager-height: #{tokens.$space-5};
+    --ids-pager-line-height: #{tokens.$line-height-xxxl};
 
-  /* Item */
-  --ids-pager-item-padding: #{tokens.$space-1};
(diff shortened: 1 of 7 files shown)


packages/Pager/src/pager.scss
franckgaudin, line 1: The focus state lacks a background. If you look in Figma, you'll see that it's the same as the hover state.

pager-focus-state

packages/Pager/src/pager.scss
```diff
@@ -1,192 +1,292 @@
...
+    --ids-pager-item-wrapper-border-size: 0.0625rem;
+    --ids-pager-item-wrapper-padding: calc(var(--hop-space-inset-xs) - var(--ids-pager-item-wrapper-border-size));
+    --ids-pager-item-wrapper-radius: var(--hop-shape-rounded-md);
+    --ids-pager-item-padding: 0;
+    --ids-pager-item-background: var(--hop-neutral-surface);
+    --ids-pager-item-border: none;
+    --ids-pager-item-prev-border-radius: var(--hop-shape-rounded-md);
+    --ids-pager-item-next-border-radius: var(--hop-shape-rounded-md);
+    --ids-pager-item-color: var(--hop-neutral-text);
+    --ids-pager-item-radius: 999px;
```
franckgaudin, line 78: Use the hopper tokens 
```suggestion
    --ids-pager-item-radius: var(--hop-shape-circle);
```

packages/Pager/src/pager.scss
```diff
@@ -1,192 +1,292 @@
...
+    --ids-pager-item-selected-background: var(--hop-neutral-surface-strong);
+    --ids-pager-item-selected-color: var(--hop-neutral-text-strong);
+
+    /* Hover */
+    --ids-pager-item-hover-border: none;
+    --ids-pager-item-hover-background: var(--hop-neutral-surface-hover);
+
+    /* Focus */
+    --ids-pager-item-focus-border-width: 0;
+    --ids-pager-item-focus-outline-width: 0.09375rem;
```
franckgaudin, line 94: There is an eslint rule that limits the number after the decimal point, use `0.0938rem`

fraincs: That would not equals 1.5px. Base 16 uses a lot of decimals versus base 10. As in the end screens renders px's I would rethink this rule, what is the goal of that rule?

franckgaudin: This rule has been defined in the standards of the [wl-web-config](https://github.com/gsoft-inc/wl-web-configs) library, so you can open an issue to have it changed.

franckgaudin (approved):

ofrogon (approved):

PR #532, opened 2023-09-14
Feature/toaster

fraincs:
Rebrand of the Toaster component.

### DIFF
--- a/packages/Toaster/src/Toast.tsx
+++ b/packages/Toaster/src/Toast.tsx
@@ -85,6 +85,7 @@ const Toast: React.FunctionComponent<ToastProps<ToastArgs>> = ({
             )}
             {isClosable && getBrand() === "workleap" && (
                 <button
+                    {...closeButtonProps}
                     onClick={() => {
                         state.close(toast.key);
                     }}
--- a/packages/Toaster/src/Toast.tsx
+++ b/packages/Toaster/src/Toast.tsx
@@ -30,6 +30,10 @@ const Toast: React.FunctionComponent<ToastProps<ToastArgs>> = ({
         toastRef
     );
 
+    const getBrand = (): string => {
+        return document.documentElement.getAttribute("data-brand") ?? "igloo";
+    };
+
     delete closeButtonProps.onPress;
 
     React.useEffect(() => {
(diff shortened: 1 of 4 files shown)


packages/Toaster/src/Toast.tsx
```diff
@@ -79,6 +83,16 @@ const Toast: React.FunctionComponent<ToastProps<ToastArgs>> = ({
                     icon={<Close size="medium" />}
                 />
             )}
+            {isClosable && getBrand() === "workleap" && (
+                <button
+                    onClick={() => {
+                        state.close(toast.key);
+                    }}
+                    className="ids-toast__close-button"
```
franckgaudin, line 91: I don't understand the need to change className in the context of workleap, can you tell me a bit more?

packages/Toaster/src/toaster.scss
franckgaudin, line 1: The icon isn't the same as in Figma. Have you checked with the design team that it's correct if we keep the Igloo version?

Capture d’écran, le 2023-09-14 à 15 52 59

packages/Toaster/src/Toast.tsx
fraincs: Yep as we don't have a ghost button that has no background change on hover design suggested using a simple icon in a button instead of using the Ghost Secondary button as it would look odd.

image

packages/Toaster/src/toaster.scss
fraincs: I will ask, I assumed icons were out of scope.

packages/Toaster/src/Toast.tsx
franckgaudin: perfect, in which case you're missing the button surrounding the icon.

fraincs:
```
            {isClosable && getBrand() === "workleap" && (
                <button
                    onClick={() => {
                        state.close(toast.key);
                    }}
                    className="ids-toast__close-button"
                >
                    <Close size="medium" />
                </button>
            )}
```

What do you mean?

packages/Toaster/src/toaster.scss
fraincs: These are out of scope according to design.

packages/Toaster/src/Toast.tsx
franckgaudin: Sorry my bad, 

I notice that the `{...closeButtonProps}` props aren't passed in the case of rebrand - is this intentional?

fraincs: my bad fixed

franckgaudin (approved):

PR #545, opened 2023-09-19
Rebrand Select component the new brand

franckgaudin:
Ticket: IGL-65

### DIFF
--- a/packages/Select/src/select-value.scss
+++ b/packages/Select/src/select-value.scss
@@ -1,56 +1,71 @@
-@use '~@igloo-ui/tokens/dist/base10/variables' as tokens;
+@use "~@igloo-ui/tokens/dist/base10/variables" as tokens;
 
 :root {
-  /* Default */
-  --ids-select-value-padding: #{tokens.$space-2} #{tokens.$space-3};
+    /* Default */
+    --ids-select-value-padding: #{tokens.$space-2} #{tokens.$space-3};
 
-  /* Compact **/
-  --ids-select-value-padding-compact: #{tokens.$space-1} #{tokens.$space-3};
+    /* Compact **/
+    --ids-select-value-padding-compact: #{tokens.$space-1} #{tokens.$space-3};
 
-  /* Disabled */
-  --ids-select-value-color-disabled: #{tokens.$grey-500};
+    /* Disabled */
+    --ids-select-value-color-disabled: #{tokens.$grey-500};
 
-  /* Placeholder */
-  --ids-select-value-color-placeholder: #{tokens.$grey-400};
+    /* Placeholder */
+    --ids-select-value-color-placeholder: #{tokens.$grey-400};
 
-  /* Visual Identifier */
-  --ids-select-value-icon-size: #{tokens.$space-4};
-  --ids-select-value-icon-margin-right: #{tokens.$space-2};
+    /* Visual Identifier */
+    --ids-select-value-icon-size: #{tokens.$space-4};
+    --ids-select-value-icon-margin-right: #{tokens.$space-2};
+}
+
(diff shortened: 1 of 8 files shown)


packages/Select/src/select-value.scss
```diff
@@ -1,56 +1,71 @@
...
+
+    /* Compact **/
+    --ids-select-value-padding-compact: 0.375rem var(--hop-space-inline-md);
+
+    /* Placeholder */
+    --ids-select-value-color-placeholder: var(--hop-neutral-text-weakest);
+
+    /* Visual Identifier */
+    --ids-select-value-icon-size: 1.5rem;
+    --ids-select-value-icon-margin-right: 0.5rem;
```
fraincs, line 33: This could be a token.

packages/Select/src/select-value.scss
```diff
@@ -1,56 +1,71 @@
...
+    --ids-select-value-icon-size: #{tokens.$space-4};
+    --ids-select-value-icon-margin-right: #{tokens.$space-2};
+}
+
+[data-brand="workleap"] {
+    /* Default */
+    --ids-select-value-padding: var(--hop-space-stack-sm) var(--hop-space-inline-md);
+
+    /* Compact **/
+    --ids-select-value-padding-compact: 0.375rem var(--hop-space-inline-md);
```
fraincs, line 26: If the padding is 6 because of the border (2 x 1px) this could use the space-inline-md - border-width

packages/Select/src/select.scss
```diff
@@ -1,93 +1,142 @@
...
+
+    /* Hover */
+    --ids-select-background-hover: var(--hop-neutral-surface-hover);
+    --ids-select-border-color-hover: var(--hop-neutral-border-hover);
+
+    /* Active */
+    --ids-select-border-color-active: var(--hop-neutral-border-active);
+
+    /* Focus */
+    --ids-select-background-focus:  var(--hop-neutral-surface-hover);
```
fraincs, line 69: double space here

packages/Select/src/select.scss
```diff
@@ -1,93 +1,142 @@
...
-    border-color: var(--ids-select-border-color-active);
-  }
-
-  &--disabled,
-  &--disabled:focus {
-    color: var(--ids-select-color-disabled);
-    background-color: var(--ids-select-background-disabled);
-    border-color: var(--ids-select-border-color-disabled);
-    cursor: not-allowed;
+    transition: border-color .2s ease 0s;
```
fraincs, line 95: whats the use of the 0s here?

alexasselin008 (dismissed):

packages/Select/src/select.scss
franckgaudin: it's not a change I've made, I don't know the reason for such timing

alexasselin008 (approved):

packages/Select/src/select-value.scss
```diff
@@ -23,14 +23,14 @@
     --ids-select-value-padding: var(--hop-space-stack-sm) var(--hop-space-inline-md);
 
     /* Compact **/
-    --ids-select-value-padding-compact: 0.375rem var(--hop-space-inline-md);
+    --ids-select-value-padding-compact: calc(var(--hop-space-stack-sm) - var(--ids-select-broder-width) * 2) var(--hop-space-inline-md);
```
fraincs, line 26: `broder-width ` should be `border-width`

fraincs (changes requested):

packages/Select/src/select-value.scss
franckgaudin: good catch

fraincs (approved):

PR #554, opened 2023-09-22
Rebrand text editor and Fix character limit

franckgaudin:
Stories IGL-11 IGL-69

### DIFF
--- a/packages/TextEditor/src/plugins/toolbar.scss
+++ b/packages/TextEditor/src/plugins/toolbar.scss
@@ -30,7 +30,7 @@
     --ids-toolbar-font-size: var(--hop-body-sm-font-size);
     --ids-toolbar-border-color: var(--hop-neutral-border-weak);
     --ids-toolbar-border-width: 0.0625rem;
-    --ids-toolbar-padding: var(--hop-space-stack-xs) var(--hop-space-inline-md);
+    --ids-toolbar-padding: var(--hop-space-inset-xs) var(--hop-space-inset-md);
 
     /* Disabled */
     --ids-toolbar-border-color-disabled: var(--hop-neutral-border-disabled);
--- a/packages/TextEditor/src/text-editor.scss
+++ b/packages/TextEditor/src/text-editor.scss
@@ -107,7 +107,7 @@
     --ids-text-editor-border-color: var(--hop-neutral-border);
     --ids-text-editor-border-radius: var(--hop-shape-rounded-md);
     --ids-text-editor-color: var(--hop-neutral-color);
-    --ids-text-editor-padding: var(--hop-space-stack-sm) var(--hop-space-inline-md);
+    --ids-text-editor-padding: var(--hop-space-inset-sm) var(--hop-space-inset-md);
     --ids-text-editor-line-height: var(--hop-body-md-line-height);
     --ids-text-editor-placeholder-color: var(--hop-neutral-text-weakest);
     --ids-text-editor-min-height: 7.5rem;
@@ -156,7 +156,7 @@
     --ids-text-editor-code-font-size: var(--hop-body-sm-font-size);
     --ids-text-editor-code-line-height: var(--hop-body-sm-line-height);
     --ids-text-editor-code-margin: var(--hop-space-stack-sm) 0;
-    --ids-text-editor-code-padding: var(--hop-space-stack-sm) var(--hop-space-inline-sm) var(--hop-space-stack-sm) 3rem;
+    --ids-text-editor-code-padding: var(--hop-space-inset-sm) var(--hop-space-inset-sm) var(--hop-space-inset-sm) 3rem;
     --ids-text-editor-code-before-background: var(--hop-neutral-surface-weak);
     --ids-text-editor-code-before-border-color: var(--hop-neutral-border-weak);
     --ids-text-editor-code-before-color: var(--hop-neutral-text);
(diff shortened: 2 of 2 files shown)


packages/TextEditor/src/plugins/toolbar.scss
```diff
@@ -2,72 +2,106 @@
...
+    --ids-toolbar-divider-width: 0.1rem;
+}
+
+[data-brand="workleap"] {
+    /* Default */
+    --ids-toolbar-font-family: var(--hop-body-sm-font-family);
+    --ids-toolbar-font-size: var(--hop-body-sm-font-size);
+    --ids-toolbar-border-color: var(--hop-neutral-border-weak);
+    --ids-toolbar-border-width: 0.0625rem;
+    --ids-toolbar-padding: var(--hop-space-stack-xs) var(--hop-space-inline-md);
```
fraincs, line 33: stack and inline are for margins only

packages/TextEditor/src/plugins/toolbar.scss
```diff
@@ -2,72 +2,106 @@
...
+
+    /* Disabled */
+    --ids-toolbar-border-color-disabled: var(--hop-neutral-border-weak);
```
fraincs, line 36: Usually a disabled border should be --hop-neutral-border-disabled, maybe check with design.

packages/TextEditor/src/text-editor.scss
```diff
@@ -1,381 +1,493 @@
...
+[data-brand="workleap"] {
+    /* === Default === */
+    --ids-text-editor-font-family: var(--hop-body-md-font-family);
+    --ids-text-editor-font-size: var(--hop-body-md-font-size);
+    --ids-text-editor-background: var(--hop-neutral-surface);
+    --ids-text-editor-border-width: 0.0625rem;
+    --ids-text-editor-border-color: var(--hop-neutral-border);
+    --ids-text-editor-border-radius: var(--hop-shape-rounded-md);
+    --ids-text-editor-color: var(--hop-neutral-color);
+    --ids-text-editor-padding: var(--hop-space-stack-sm) var(--hop-space-inline-md);
```
fraincs, line 108: stack and inline is for margin

packages/TextEditor/src/text-editor.scss
```diff
@@ -1,381 +1,493 @@
...
+    /* Link */
+    --ids-text-editor-link-color: var(--hop-primary-text);
+    --ids-text-editor-link-font-weight: var(--hop-body-sm-font-weight);
+    --ids-text-editor-link-hover: var(--hop-primary-text-hover);
+
+    /* Code */
+    --ids-text-editor-code-font-size: var(--hop-body-sm-font-size);
+    --ids-text-editor-code-line-height: var(--hop-body-sm-line-height);
+    --ids-text-editor-code-margin: var(--hop-space-stack-sm) 0;
+    --ids-text-editor-code-padding: var(--hop-space-stack-sm) var(--hop-space-inline-sm) var(--hop-space-stack-sm) 3rem;
```
fraincs, line 155: inline and stack are for margins

packages/TextEditor/src/text-editor.scss
```diff
@@ -1,381 +1,493 @@
...
+    --ids-text-editor-heading-2-font-size: var(--hop-heading-lg-font-size);
+    --ids-text-editor-heading-2-font-weight: var(--hop-heading-lg-font-weight);
+    --ids-text-editor-heading-2-margin: 0 0 var(--hop-space-stack-sm);
+
+    /* Quote */
+    --ids-text-editor-quote-color: var(--hop-neutral-text);
+    --ids-text-editor-quote-border-color: var(--hop-neutral-border);
+    --ids-text-editor-quote-border-size: 0.25rem;
+    --ids-text-editor-quote-margin: 0 0 var(--hop-space-stack-sm) var(--hop-space-inline-md);
+    --ids-text-editor-quote-padding: 0 0 0 var(--hop-space-inline-md);
```
fraincs, line 188: inline is for margin

packages/TextEditor/src/plugins/toolbar.scss
franckgaudin: It's strange to have tokens only for one use case, what's the reason for this? In this case, what values should be used?

packages/TextEditor/src/plugins/toolbar.scss
franckgaudin: The question has been put to the Design team, and I'm waiting for an answer.

alexasselin008 (dismissed):

packages/TextEditor/src/plugins/toolbar.scss
franckgaudin: Done

packages/TextEditor/src/plugins/toolbar.scss
fraincs: > I've validated this in other components and we've already used these tokens as padding values in the past.
This might have been an oversight.

Stacking is a form of spacing that creates equal distance between components or a group of items.

In your situation here you would need : `--hop-space-inset-xs --hop-space-inset-md` although we could challenge design that they should have used `--hop-space-inset-squish-md`

franckgaudin: I understand, thank you for the clarification.

In my opinion the documentation is not clear enough on which `space-` tokens to use, we should improve this in a future iteration. When do you say?

fraincs (approved):

alexasselin008 (approved):

Issue #541, opened 2023-09-18
[Bug]: It is possible to add more than the maxTags tags in the TagPicker


sbaboot:
### Contact Details

t-sebastien.flamant@workleap.com

### What happened?

When we already have selectedResults (so the first render of the TagPicker has selectedResults), it is possible to add more tags in the TagPicker than the maxTags prop. For exemple, I want a max of 5 tags. When my TagPicker is empty at the first render, I can put max 5 tags. But when my selectedResults contain already 2 tags at the first render, I can now have 7 tags (and I don't want it).
![tag picker max](https://github.com/gsoft-inc/ov-igloo-ui/assets/45638578/eee2069f-895a-4dda-8cb2-5c72d016f494)

### Component

TagPicker

### Component Version

0.5.0

PR #540, opened 2023-09-18
Feature/datepicker

fraincs:
Rebrand of DatePicker - Ticket : IGL-70

PR #555, opened 2023-09-25
Feature/tagpicker

fraincs:
IGL-68 / IGL-87

Rebrand of TagPicker / Fix for #541


alexasselin008:
you should link this PR to opened bugs in this repo, so they close once it merges

packages/TagPicker/src/TagPicker.tsx
```diff
@@ -283,6 +283,14 @@ const TagPicker: React.FunctionComponent<TagPickerProps> = ({
         }
     }, [tagRemoved, inputDisabled, handleGainFocus, autoFocus]);
 
+    useEffect(() => {
```
alexasselin008, line 286: What is the reason of this useEffect? It feels like something that we could handle in an action handler instead of a use effect. 

Why is the check in handleResultSelection not enough?

packages/TagPicker/src/TagPicker.tsx
```diff
@@ -184,7 +184,7 @@ const TagPicker: React.FunctionComponent<TagPickerProps> = ({
         onSelection(resultId);
         resetSearch();
 
-        if (maxTags && selectedResultsCount === maxTags - 1) {
+        if (maxTags && selectedResultsCount >= maxTags - 1) {
```
alexasselin008, line 187: thats an odd change to the check. 

before: "If there is only 1 tag spot left, then do this"
now: "If there are more tags than allowed, then do this". Why is this possible? 

It feels like treating the symptoms more than the cause of the issue

packages/TagPicker/src/TagPicker.tsx
fraincs: Since on load you could already have selected tags, this assures that the user does not try to pick a tag when the limit is already reached, see issue https://github.com/gsoft-inc/ov-igloo-ui/pull/540

packages/TagPicker/src/TagPicker.tsx
fraincs: Indeed, upon further exploration the -1 does not seems to make sense at all. As for the > this was pretty much a defensive way of making sure that if we somehow have more than the max it will be flagged.

packages/TagPicker/src/TagPicker.tsx
```diff
@@ -118,12 +118,23 @@ const TagPicker: React.FunctionComponent<TagPickerProps> = ({
     const [keyboardFocusIndex, setKeyboardFocusIndex] = useState(
         defaultKeyboardFocusIndex
     );
+    const selectedResults = maxTags ? selectedResultsProp.slice(0, maxTags) : selectedResultsProp;
     const selectedResultsCount = selectedResults.length;
+    const hasTooManyTags = selectedResults.length !== selectedResultsProp.length;
+
+    hasTooManyTags && console.warn(
```
franckgaudin, line 125: Is it necessary to display a warning in the console for this use case, as the information is already being returned to the user?

packages/TagPicker/src/TagPicker.tsx
```diff
@@ -297,7 +316,8 @@ const TagPicker: React.FunctionComponent<TagPickerProps> = ({
                 color={s.color}
                 icon={s.icon}
                 dismissible={!disabled}
-                appearance="neutral"
+                appearance={getBrand() === "workleap" ? "default" : "neutral"}
+                rounded={getBrand() === "workleap"}
```
franckgaudin, line 320: With version 1.2.0 of the `Tag` component, this declaration is no longer necessary, since by default they are round.

packages/TagPicker/src/tag-picker.scss
franckgaudin, line 1: A little space seems to be missing to the left of the icon.

tag-picker-search-icon

alexasselin008 (approved):

franckgaudin (approved):

PR #571, opened 2023-10-31
fix for IGL 93

fraincs:
- Icon in Alert will only show in "desktop" mode.
- Close button will now always show if the closable prop is true.

### DIFF
--- a/packages/Alert/src/alert.scss
+++ b/packages/Alert/src/alert.scss
@@ -157,8 +157,8 @@
                 var(--ids-alert-border-color-announcement),
                 var(--ids-alert-color-announcement),
                 var(--ids-alert-background-color-announcement),
-        var(--ids-alert-icon-color-announcement)
-      );
+                var(--ids-alert-icon-color-announcement)
+            );
         }
 
         &--info {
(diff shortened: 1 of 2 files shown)


packages/Alert/src/alert.scss
```diff
@@ -261,13 +255,14 @@
...
     &__icon {
         font-size: 0;
+        display: none;
 
         + .ids-alert__body {
-            margin-left: var(--ids-alert-flex-gap-component);
+            margin-left: 0;
         }
 
         @media (width >= #{tokens.$breakpoints-sm}) {
```
vicky-comeau, line 264: I'm guessing when I updated the way the breakpoints were written, this used to be <= and I wrote it wrong, but the way you did it works too.

packages/Alert/src/alert.scss
```diff
@@ -261,13 +255,14 @@
 
     &__icon {
         font-size: 0;
+        display: none;
 
         + .ids-alert__body {
-            margin-left: var(--ids-alert-flex-gap-component);
+            margin-left: 0;
```
vicky-comeau, line 261: why is this being replaced with 0 instead of using the variable?

packages/Alert/src/alert.scss
fraincs: I can also only flip the > to <. I tend to go mobile first.

vicky-comeau: yes mobile first is the way to go. Some were just done the quick way I think... Not a huge deal in this case.

vicky-comeau (approved):

franckgaudin (approved):

PR #573, opened 2023-11-03
Feat: [IGL-95] Filter rebrand

vicky-comeau:


.storybook/useBrand.js
```diff
@@ -2,62 +2,10 @@ import { useEffect } from 'react';
 
 export const useBrand = (StoryFn, context) => {
     const {brand} = context.globals;
-    const {displayName} = context.component;
-    const components = [
```
fraincs, line 6: I'm out of the loop here, what's the goal of this removal?

packages/Color/src/Color.tsx
```diff
@@ -1,81 +1,99 @@
...
-          break;
+    const classes = cx("ids-color", `ids-color--${size}`, className);
+
+    const initialsFromName = React.useMemo(() => {
+        if (!name) {
+            return null;
+        }
+
+        let initials = "";
+        const splitName = name.trim().split(" ");
```
fraincs, line 48: If the name has more than one space this won't work.

packages/ColorPicker/src/ColorPicker.tsx
```diff
@@ -14,26 +14,160 @@ import ColorPickerOption from "./ColorPickerOption";
...
+    { 
+        id: "electricBlue700", 
+        value: variables.electricBlue700, 
+        hoverColor: variables.electricBlue700, 
+        iconColor: variables.samoyed 
+    }
 ] as const;
 
-export type ColorName = typeof colorNames[number]["id"];
+export const colorNamesWL = [
```
fraincs, line 111: I would use Workleap here as we are not using a shorter name for Igloo.

packages/ColorPicker/src/ColorPicker.tsx
```diff
@@ -47,15 +181,36 @@ export interface ColorPickerProps {
...
+    const brand = propBrand ?? document.documentElement.getAttribute("data-brand") ?? "igloo";
+    
+    const getDefaultColor = (): ColorName => {
+        return brand === "workleap" ? "decorativeOption3" : "dandelion200";
+    };
+
+    const selectedColor = propSelectedColor ?? getDefaultColor();
+
+    let colorNames: ReadonlyArray<Readonly<ColorNameItem>>;
+    if (brand === "workleap") {
```
fraincs, line 200: I would add a return before the if

packages/ColorPicker/src/__snapshots__/ColorPicker.test.tsx.snap
```diff
@@ -9,11 +9,13 @@ exports[`ColorPicker It should render a snapshot 1`] = `
     <button
       class="ids-color-picker__button"
       data-color="dandelion200"
+      data-rac=""
```
fraincs, line 12: What is it used for?

packages/ColorPicker/src/color-picker-option.scss
```diff
@@ -15,9 +15,23 @@
...
-  
-  /* Icon */
-  --ids-color-picker-option-icon-color: #{tokens.$samoyed};
+
+}
+
+[data-brand="workleap"] {
+  /* Default */
+  --ids-color-picker-option-border-radius: var(--hop-shape-circle);
+  --ids-color-picker-option-border-size: 0.09rem;
```
fraincs, line 24: 1.44px seems arbitrary.

packages/ColorPicker/src/color-picker.scss
```diff
@@ -12,13 +15,33 @@
...
   --ids-color-picker-btn-border-size: 0.15rem;
-  --ids-color-picker-btn-focus: #{tokens.$focus};
+  --ids-color-picker-btn-focus: #{tokens.$electric-blue-600};
   --ids-color-picker-btn-padding: #{tokens.$space-2};
 }
 
+[data-brand="workleap"] {
+  /* Default */
+  --ids-color-picker-font-family: var(--hop-body-sm-font-family);
+  --ids-color-picker-font-size: var(--hop-body-sm-font-size);
```
fraincs, line 29: Font weight and line-height are missing.

.storybook/useBrand.js
vicky-comeau: We did the rebrand everywhere, so we don't need to specify specific components anymore.

packages/ColorPicker/src/ColorPicker.tsx
vicky-comeau: not sure I understand, a return for what?

packages/ColorPicker/src/color-picker-option.scss
vicky-comeau: Well it called for 1.5px so I rounded 0.09375rem. That's just way too long. Plus browsers don't end up computing the exact amount anyway.

packages/ColorPicker/src/__snapshots__/ColorPicker.test.tsx.snap
vicky-comeau: This is a react-aria button so I'm guessing it's added by them. I double checked the code, not something we added.

alexasselin008: react-aria-components (rac) seems to be adding this property on react aria components. 

There is a major release releasing todayt of react-aria-components, and the code that adds the data-rac is still there. I think it's used by their tailwindcss integration to target react-aria-components. 

https://github.com/adobe/react-spectrum/blob/9ce2f674eab2cc8912800d3162dcf00a1ce94274/packages/react-aria-components/src/utils.tsx#L134-L135

alexasselin008: in all cases, don't worry about it for now i think

packages/ColorPicker/src/color-picker-option.scss
fraincs: True in that case we might need to ask design to stop using .5px for borders 😄

packages/ColorPicker/src/ColorPicker.tsx
fraincs: new line

packages/ColorPicker/src/color-picker-option.scss
vicky-comeau: haha maybe... I guess I'll keep it for now?

fraincs (approved):

franckgaudin (approved):

fraincs (approved):

Issue #593, opened 2023-12-11
[Bug]: Select & Combobox - Make sure the focus state is the same as Orbiter's focus state


vicky-comeau:
### What happened?

Igloo's select should have the focus state when in focus, but when the select is open, the focus is on the dropdown so the trigger should only have the active look, not the focus look.

### Component

Select, Combobox

### Component Version

latest

PR #594, opened 2023-12-11
Select & Combobox - Make sure the focus state is the same as Orbiter's focus state

fraincs:
Fix for issue #593 : Select & Combobox - Make sure the focus state is the same as Orbiter's focus state

### DIFF
--- a/packages/Combobox/src/combobox.scss
+++ b/packages/Combobox/src/combobox.scss
@@ -137,6 +137,7 @@
         &:focus-visible,
         &.focus {
             background-color: var(--ids-combobox-background-active);
+            border-color: var(--ids-select-border-color-active);
             box-shadow: none;
         }
     }
--- a/packages/Combobox/src/combobox.scss
+++ b/packages/Combobox/src/combobox.scss
@@ -115,7 +115,7 @@
         background-color: var(--ids-combobox-background-hover);
     }
 
-    &:focus,
+    &:focus-visible,
     &.focus {
         outline: 0;
         box-shadow: var(--ids-combobox-focus);
@@ -131,6 +131,11 @@
     &--active,
     &.active {
         border-color: var(--ids-combobox-border-color-active);
+
+        &:focus,
+        &.focus {
+            box-shadow: none;
+        }
     }
 
     &--disabled,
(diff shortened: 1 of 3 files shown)


packages/Combobox/src/combobox.scss
```diff
@@ -131,6 +131,11 @@
     &--active,
     &.active {
         border-color: var(--ids-combobox-border-color-active);
+
+        &:focus,
```
vicky-comeau, line 135: I don't think this is supposed to be inside --active class

packages/Select/src/select.scss
```diff
@@ -122,6 +122,11 @@
     &--active,
     &.active {
         border-color: var(--ids-select-border-color-active);
+
+        &:focus,
```
vicky-comeau, line 126: same as above

packages/Combobox/src/combobox.scss
```diff
@@ -131,6 +131,11 @@
     &--active,
     &.active {
```
vicky-comeau, line 132: I think we need to specify a white background color for active. If not, we end up with a grey background when the menu is open:
image
This is taken for the select, but it's the same for the combobox.

packages/Combobox/src/combobox.scss
fraincs: Opening a Select/Combobox gives the select both the active and focus class. This is why this is inside active.

packages/Select/src/select.scss
```diff
@@ -106,22 +113,27 @@
...
     }
 
     &--active,
     &.active {
+        background-color: var(--ids-select-background-active);
         border-color: var(--ids-select-border-color-active);
+
+        &:focus-visible,
+        &.focus {
+            background-color: var(--ids-select-background-active);
```
vicky-comeau, line 134: is it showing the active border color when it has focus?

vicky-comeau: doesn't look like it in the preview, I'd add the active border color here too. (Combobox too)

vicky-comeau (approved):

franckgaudin (approved):

PR #609, opened 2023-12-15
Fix component rendering when used with a tooltip

franckgaudin:
add a data tag to the first and last elements to style borders radius

### DIFF
--- a/packages/ButtonGroup/src/button-group.scss
+++ b/packages/ButtonGroup/src/button-group.scss
@@ -5,233 +5,242 @@
 @import url('~@hopper-ui/tokens/tokens.css');
 
 :root {
-  /* Default */
-  --ids-button-group-font-family: #{tokens.$primary-font-family};
-  --ids-button-group-font-size: #{tokens.$font-size-4};
-  --ids-button-group-font-weight: #{tokens.$font-weight-regular};
-  --ids-button-group-text: #{tokens.$grey-800};
-  --ids-button-group-box-sizing: content-box;
-  --ids-button-group-display: flex;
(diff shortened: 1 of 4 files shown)


alexasselin008 (approved):

packages/ButtonGroup/src/button-group.scss
```diff
@@ -5,233 +5,242 @@
...
+    }
+
+    &:active,
+    &:hover,
+    &:focus,
+    &.is-selected {
+        z-index: var(--ids-button-group-item-index-max);
+    }
+
+    &[data-item="is-first"] {
```
vicky-comeau, line 175: I guess I don't really understand how this works.. Wouldn't this apply the border radius to the tooltips container if there is one? Or do we not care? How is this actually different from just using :first-child and :last-child?

fraincs (approved):

packages/ButtonGroup/src/button-group.scss
franckgaudin: The use of `:first-child` and `:last-child` works very well as long as the notion of `Tooltip` is not introduced. However, when the Tooltip is added, the DOM changes, so these properties no longer fulfill their role correctly.

### Example with a tooltip
```
... 
<div class="ids-btn-group">
  <span class="ids-tooltip__container">
    <button class="ids-btn-group__item">
...
```
In this example, the class `ids-btn-group__item` is both `first-child` and `last-child`, and as the last declaration in the css file is `last-child`, it applies the border radius to the right for all children. 

To counter this problem, in the parent, we convert the children into an array and apply a data tag first to the first element of the array and last to the last. In this way, we can have a slightly more precise css rule.

PR #615, opened 2023-12-19
Fix: [IGL-106] Input character count color fix and version updates

vicky-comeau:


fraincs (approved):

packages/Disclosure/src/disclosure.scss
```diff
@@ -142,6 +142,10 @@
 
 .ids-disclosure .ids-disclosure__header-icon {
     margin: var(--ids-disclosure-header-icon-margin);
+
```
alexasselin008, line 145: This change seems undocumented

.changeset/clean-needles-taste.md
```diff
@@ -0,0 +1,51 @@
...
+"@igloo-ui/text-editor": patch
+"@igloo-ui/textarea": patch
+"@igloo-ui/toaster": patch
+"@igloo-ui/toggle": patch
+"@igloo-ui/tooltip": patch
+"@igloo-ui/vertical-bar-chart": patch
+"docs": patch
+---
+
+Updated input icon and character count colors. Also updated hopper token versions and igloo icon versions.
```
alexasselin008, line 51: i think the best practice here would be to create 2 changeset

texteditor with the message: Updated input icon and character count colors
all other pacakges: updated hopper token versions and igloo icon versions.

packages/Disclosure/src/disclosure.scss
vicky-comeau: ya when I was testing I noticed that the variable wasn't written right.. should I add this to my changeset?

.changeset/clean-needles-taste.md
vicky-comeau: ok I will created two changesets, thanks

packages/Disclosure/src/disclosure.scss
vicky-comeau: I updated the changesets

alexasselin008 (approved):

fraincs (approved):

PR #620, opened 2023-12-20
Fix: [IGL-108] Fixed error styles for TagPicker

vicky-comeau:


packages/TagPicker/src/tag-picker.scss
```diff
@@ -141,10 +144,12 @@
     margin: var(--ids-tag-picker-tag-margin);
 
     .ids-tag-picker & {
+        background: var(--ids-tag-picker-tag-background);
         border: var(--ids-tag-picker-tag-border);
     }
 
-    .ids-tag-picker--error & {
+    .ids-tag-picker &--error {
```
fraincs, line 151: shouldn't this be `.ids-tag-picker--error`

vicky-comeau: nope. that was part of the issue. We want to set the tag border when the tag is in error, not when the tag picker is in error. They are set individually.

fraincs (approved):

ofrogon (approved):

PR #631, opened 2024-01-08
Fix: [IGL-113] Fixed disclosure styles

vicky-comeau:


packages/Disclosure/src/disclosure.scss
```diff
@@ -122,20 +123,58 @@
...
+
     &:hover,
-    &:focus {
-        background-color: var(--ids-disclosure-header-background-hover)
+    &--focus-visible {
+        background-color: var(--ids-disclosure-header-background-hover);
+
+        .ids-disclosure--low-contrast & {
+            color: var(--ids-disclosure-low-contrast-header-color-hover);
+            background: none;
```
fraincs, line 172: this should be `background-color: none`

fraincs: No, just a pet peeve, as it could override other background properties and have unwanted side effects.

alexasselin008 (dismissed):

packages/Disclosure/src/disclosure.scss
vicky-comeau: alright, updated.

alexasselin008 (approved):

fraincs (approved):

PR #648, opened 2024-01-19
Fix [IGL-117] Added missing ref & inputValue props to TagPicker

vicky-comeau:


packages/TagPicker/src/TagPicker.tsx
```diff
@@ -294,6 +302,12 @@ const TagPicker: React.FunctionComponent<TagPickerProps> = ({
         }
     }, [tagRemoved, inputDisabled, handleGainFocus, autoFocus]);
 
+    React.useEffect(() => {
```
alexasselin008, line 305: inputValue should be set in the local use state declaration instead of a the useEffect

```    
const [localInputValue, setLocalInputValue] = useState(inputValue);
```

2 main advantages: 
- less code
- At the moment, the first render has a localInputValue of "", and the useEffect is ran after, setting the state and re-rendering the component. With the default value, the first render already has the value

vicky-comeau: true that it should be set there, but I need the useEffect so that the user can update the inputValue later on and it would be synced with the localinputValue.
I couldn't figure out any other way to do this.. OV needs a way to set the input value to "". If it's not done by using state, the width of the input would not be updated.
I didn't want the user to always have to set a state variable and I wanted the TagPicker to still manage the input value state

alexasselin008 (approved):

fraincs (approved):

PR #661, opened 2024-01-29
Fix: [IGL-118] Modal & Dialog header style

vicky-comeau:


packages/Modal/src/Modal.tsx
```diff
@@ -209,12 +208,12 @@ const Modal: React.FunctionComponent<ModalProps> = (props: ModalProps) => {
...
 
                                     <IconButton
-                                        size="small"
+                                        size="medium"
                                         className="ids-modal__close"
                                         onClick={onClose}
                                         appearance={{ type: "ghost", variant: "secondary" }}
                                         aria-label={closeBtnAriaLabel}
-                                        icon={<Close />}
+                                        icon={<DismissIcon />}
```
alexasselin008, line 216: In theorie these icons should only be used when the theme is workleap. Does it affect anything here?

packages/Modal/src/Modal.tsx
```diff
@@ -192,15 +192,14 @@ const Modal: React.FunctionComponent<ModalProps> = (props: ModalProps) => {
...
-                                            size="small"
                                             className="ids-modal__back"
                                             onClick={() => {
                                                 if (carousel && carousel.currentSlide) {
                                                     handleOnPageChange(carousel.currentSlide - 1);
                                                 }
                                             }}
                                             appearance={{ type: "ghost", variant: "secondary" }}
-                                            icon={<ChevronLeft size="medium" />}
+                                            icon={<AngleLeftIcon />}
```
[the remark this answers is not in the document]
vicky-comeau: they are basically the same thing. I figured it doesn't really matter.. They should be the same size. We have no way of knowing which is Igloo and which is workleap theme inside the component unless we pass a prop, I figured it would be way more work for nothing. No one would really notice.

packages/Modal/src/Modal.tsx
vicky-comeau: Same comment as above here too :P

alexasselin008 (approved):

PR #665, opened 2024-01-30
Fix: IGL-119 Updated alert rich icons & announcement color

vicky-comeau:


packages/Alert/src/hooks/useBrand.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+        observer.observe(document.documentElement, { attributes: true });
+
+        return () => {
+            observer.disconnect();
+        };
+    }, []);
+
+    return isWorkleap;
+};
```
alexasselin008, line 24: everywhere else in the codebase, the way it's done is with : 
```
    const getBrand = (): string => {
        return document.documentElement.getAttribute("data-brand") ?? "igloo";
    };
```

why is this one different?

vicky-comeau: This is iffy and might not always work. The data-attribute is set by storybook when the global variable is updated. For a basic switch, it might work, but this hook I created watches the actual data-attribute. I feel more comfortable with this, especially since we're not testing with OV here.

alexasselin008 (approved):

PR #692, opened 2024-02-19
UI adjustment to Alert and Select

fraincs:
- Inline Alert border-radius was set at 0, where it should have been 8px.
- Icons in a "closed" Select were not the right colour.

### DIFF
--- a/packages/Alert/src/alert.scss
+++ b/packages/Alert/src/alert.scss
@@ -96,6 +96,7 @@
 
     .ids-alert {
         align-items: flex-start;
+        border-radius: var(--ids-alert-border-radius);
         display: flex;
         border-style: solid;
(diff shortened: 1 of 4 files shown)


packages/Alert/src/alert.scss
```diff
@@ -112,7 +113,6 @@
         /* Styles */
         &--card {
             background-color: var(--ids-alert-background-color-light);
-            border-radius: var(--ids-alert-border-radius);
```
vicky-comeau, line 115: does moving this do unwanted things to the old style alert?

fraincs: The old should have border radius according to KC, he went back in the Figma as fas as possible and it was an issue.

vicky-comeau (approved):

PR #701, opened 2024-02-20
IGL-122: Fixed localization in Pager. Added Provider. Updated depende…

vicky-comeau:
…ncies


packages/Provider/jest.config.js
```diff
@@ -0,0 +1,8 @@
+const base = require('../../jest.config');
+const { name } = require('./package.json');
+
+module.exports = {
+  ...base,
+  name,
+  displayName: name,
```
fraincs, line 7: we can remove the trailing comma here

fraincs (dismissed):

packages/Datepicker/src/Datepicker.tsx
```diff
@@ -21,7 +21,7 @@ import Calendar from "./components/Calendar";
 
 import "./datepicker.scss";
 
-interface Date { utc: string; local: string }
+interface Date { utc: string; local: string | null }
```
alexasselin008, line 24: local: string | null

is that related to the provider or a package update?

packages/Pager/src/Pager.tsx
```diff
@@ -153,7 +159,7 @@ const Pager: React.FunctionComponent<PagerProps> = ({
             </nav>
             <div className="ids-pager__results">
                 {pageSize * currentPage - pageSize + 1}-
-                {Math.min(pageSize * currentPage, totalCount)} of {totalCount}
+                {Math.min(pageSize * currentPage, totalCount)} {stringFormatter.format("of")} {totalCount}
```
alexasselin008, line 162: i'm 90% certain that there is a way for the stringFormatter to be used with templates. Looking at react-spectrum, it should be 

in intl file:
```
  "pageRange": "{count} of {totalCount}"
```

in js
```
stringFormatter.format('pageRange', {count: 1, totalCount: 4});
```

alexasselin008: same comment for                      

                          : `${stringFormatter.format("goToPage")} ${pageNumber}`

packages/Datepicker/src/Datepicker.tsx
vicky-comeau: It's most likely related to a package update. It complained without it.

packages/Pager/src/Pager.tsx
alexasselin008, line 162: Might not be that bad in this scenario, but i think we try to offer translation for aria-labels AND offer defaults that make sense for string that are visually rendered. But the consumers should always be able to override the strings. 

for instance, in this case, a pager should offer a method that takes 3 parameters and return a string or ReactNode, and we offer the default value which would be the string template

vicky-comeau: yes it can be, but I figured they were positioned in the same place so I didn't add it. If you think I should, then I'll fix it.

alexasselin008: React-spectrum for instance offer a "valueLabel" props

https://react-spectrum.adobe.com/react-aria/Meter.html#props

valueLabel - ReactNode- The content to display as the value's label (e.g. 1 of 4).
showValueLabel - boolean- Whether the value's label is displayed. True by default if there's a label, false by default if not.

vicky-comeau: so from what I understand, you want me to add a prop like renderResultsText that passes the pageSize, currentPage and totalCount?
If not set, it defaults to our default message?

packages/Provider/src/Provider.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from "react";
+
+import { I18nProvider, type I18nProviderProps, useLocale, useLocalizedStringFormatter } from "react-aria";
+
+import "./provider.scss";
```
alexasselin008, line 5: remove this line, there is no scss

packages/Provider/src/intl/en-US.json
```diff
@@ -0,0 +1,3 @@
+{
+    "testMessage": "test message"
```
alexasselin008, line 2: Hmm i don't like the idea of having localized test messages. those folders are being shipped to npm, and if we were to add non-test strings, then both production and tests strings would be loaded by client applications

packages/Provider/src/Provider.tsx
vicky-comeau: I originally had done this, but it kept complaining that there was no provider.css in the dist file... I removed all references to it as far as I know, it might have been because we use it in the decorator..

packages/Provider/src/intl/en-US.json
alexasselin008, line 2: I wouldn't add test messages in the intl folder. We might end up with a mix and match of real and tests values, and they are all going to be sent to the users. I suggest you create a local object in your test file instead

packages/Provider/src/Provider.tsx
alexasselin008: and delete the .scss file as well. No need to keep stuff just in case

packages/Pager/package.json
```diff
@@ -24,8 +24,9 @@
     "build": "rollup -c rollup.config.js"
   },
   "dependencies": {
+    "@igloo-ui/provider": "^0.0.0",
```
alexasselin008, line 27: it has to be a peer dependency and dev dependency

packages/Pager/src/Pager.tsx
alexasselin008: We should do it the right way ya. to start getting used to it

alexasselin008: They have all the informations already, so the function is overkill.

add to the interface
```
valueLabel?: ReactNode
```

and we set it's default value in our codebase if it's null
```
valueLabel = stringFormatter.format("pageRange", { minRange: pageSize * currentPage - pageSize + 1, maxRange: Math.min(pageSize * currentPage, totalCount), totalCount: totalCount })
```

you can chose variable name that makes more sense in the context (minRange,maxRange,totalCount)

packages/Provider/src/Provider.tsx
alexasselin008: I'll investigate, there is no reason this file should exist.

alexasselin008 (approved):

PR #703, opened 2024-02-20
StackedBarTooltip title and dataSet custom classes

ThiBsc:
# Why
We need to add a Title in the StackedBar Popover and also be able to have custom classes both on Bar and ListItem

# How
Using the DataSet data to specify a custom className for each Bar and ListItem related to the Bar

## StackedBar
image

## Tests run
```diff
+ Test Suites: 45 passed, 45 total
+ Tests:       295 passed, 295 total
+ Snapshots:   63 passed, 63 total
+ Time:        53.728 s
```


alexasselin008 (changes requested):
Poked on slack, waiting for answer

alexasselin008 (approved):

PR #705, opened 2024-02-20
Fix: [IGL-123] Centered check & switched to svgs

vicky-comeau:


vicky-comeau:
For your info, I stayed with igloo icons because the hopper icons was not using a stroke so I couldn't do the animation.

alexasselin008 (approved):

fraincs (approved):

PR #708, opened 2024-02-21
Fix: IGL-124 Select, Combobox, TagPicker, List, Tag and VisualIdentif…

vicky-comeau:
…ier style adjustments

### DIFF
--- a/packages/List/src/ListItem.tsx
+++ b/packages/List/src/ListItem.tsx
@@ -136,7 +136,7 @@ const ListItem: React.FunctionComponent<ListItemProps> = ({
         "ids-list-item--loading": loading
     });
 
-    let visualIdentifierSize = "medium";
+    let visualIdentifierSize: Size = "medium";
     if (option?.type === "member") {
         visualIdentifierSize = "large";
     } else if (isCompact) {
@@ -159,7 +159,7 @@ const ListItem: React.FunctionComponent<ListItemProps> = ({
                 icon={option?.icon}
                 color={option?.color}
                 src={option?.src}
-                size={visualIdentifierSize as Size}
+                size={visualIdentifierSize}
             />
         </div>
     );
--- a/packages/List/src/ListItem.tsx
+++ b/packages/List/src/ListItem.tsx
@@ -2,7 +2,7 @@ import * as React from "react";
 import cx from "classnames";
 
 // eslint-disable-next-line import/no-extraneous-dependencies
-import { VisualIdentifier } from "@shared/components";
+import { VisualIdentifier, type Size } from "@shared/components";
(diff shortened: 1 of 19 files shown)


packages/List/src/ListItem.tsx
```diff
@@ -136,6 +136,13 @@ const ListItem: React.FunctionComponent<ListItemProps> = ({
         "ids-list-item--loading": loading
     });
 
+    let visualIdentifierSize = "medium";
```
alexasselin008, line 139: prefer typing the variable instead of casting it later
```
let visualIdentifierSize: Size  = "medium";
...
size={visualIdentifierSize}
```

packages/Tag/src/Tag.tsx
```diff
@@ -19,9 +19,10 @@ export type AppearanceNew =
   | "positive"
   | "caution"
   | "negative"
-  | "neutral";
+  | "neutral"
+  | "select";
```
alexasselin008, line 23: whats a "select" appearance? looks like a dropdown? is there new figma files related to this?

vicky-comeau: Well I was trying to figure out the best way to handle the select tag:
image

Before we over-rid it in the tag picker but this just didn't make sense so I tried to create a new appearance. I'm not sure if there was a better way to do this. It seemed weird to change the styles based on if it had a dismiss button or not.
I know that maybe it should have been a new component, but at this point since the dismiss button and icons were already in the Tag component, I didn't think it made sense for Igloo.

Any better ideas?

alexasselin008: In Hopper, this should be 2 components like this: 
https://wl-orbiter-website.netlify.app/?path=/docs/lozenge--default-story
https://wl-orbiter-website.netlify.app/?path=/docs/tag--default-story#default

For now we'll add a variant in igloo if it's the wayyou want to proceed

shared/components/src/visual-identifier/VisualIdentifier.tsx
```diff
@@ -51,7 +57,9 @@ VisualIdentifierProps
             return <Color className={classes} color={color} size={size} />;
         }
         if (src) {
-            return <Avatar className={classes} src={src} size={size} />;
+            const avatarSize = avatarSizeMapping[size] as Size;
```
alexasselin008, line 60: this cast casts the value incorrectly. 

```
const avatarSizeMapping = {
        "small": "xsmall",
        "medium": "small",
        "large": "medium"
    } as const satisfies Record<Size, string>
```
```
            const avatarSize = avatarSizeMapping[size];
```

PR #719, opened 2024-02-23
Removed unnecessary padding on popover title and content

fraincs:
## Popover

- Removed padding on title when no close button is present.
- Removed padding on content.
- Adjusted close button position in Workleap theme.


packages/Popover/src/popover.scss
```diff
@@ -41,6 +42,7 @@
         --ids-popover-title-color: var(--hop-neutral-text);
         --ids-popover-title-font-weight: var(--hop-body-sm-semibold-font-weight);
         --ids-popover-title-padding: var(--hop-space-inset-xl);
+        --ids-popover-close-offset: 1rem;
```
vicky-comeau, line 45: couldn't you use --hop-space-inline-md or something for this?

fraincs: Yep totally will fix

vicky-comeau (approved):

Issue #716, opened 2024-02-23
[Bug]: StackedBar border radius not applied correctly for items with zero value


alexasselin008:
### What happened?

The radius in the StackedBar component is currently only being applied to the first and last items. However, this implementation leads to a scenario where, in cases of items with a value of 0, the radius is not visually represented as expected.

![image](https://github.com/gsoft-inc/ov-igloo-ui/assets/38871812/5a996ee9-4e75-4876-ad6b-53756a454744)

### Component

StackedBar

### Component Version

latest

fraincs:
Fixed in https://github.com/gsoft-inc/ov-igloo-ui/pull/725

PR #725, opened 2024-02-26
fix for border radius in the StackedBar component

fraincs:
Fix for issue #716.

The way the code was working is that it was checking for the first item in the dataSet even if it's value was 0, items with a 0 value are not rendered, resulting in situations where we had items with no border radius even though they were the first item rendered. The same applies to the lastItem.

### DIFF
--- a/packages/StackedBar/src/StackedBar.stories.tsx
+++ b/packages/StackedBar/src/StackedBar.stories.tsx
@@ -207,6 +207,47 @@ export const CustomRangeAndFormat = {
   },
 };
 
+export const ZeroFirstAndLast = {
+    args: {
+      showValue: true,
+      valueRange: { min: 0, max: 'dataMax' },
+      formatValue: (value: number) => {
+        return `${value} units`;
+      },
+      dataSet: [
+        {
+          key: 'stronglyUnfavorable',
+          label: 'Strongly Unfavorable',
+          value: 0,
+          strength: -2,
+        },
+        {
+          key: 'unfavorable',
+          label: 'Unfavorable',
+          value: 135,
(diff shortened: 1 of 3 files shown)


packages/StackedBar/src/StackedBar.stories.tsx
```diff
@@ -207,6 +207,47 @@ export const CustomRangeAndFormat = {
   },
 };
 
+export const ZeroFirstAndLast = {
```
alexasselin008, line 210: j'ai pas QA encore, mais j'ai comme le feeling que la solution que tu as supporterais pas que les 2-3 premiers soit à 0, un chiffre ds le millieu + 2-3 dernier à 0?

vicky-comeau: it doesn't work if all are 0 except for the last:
image

vicky-comeau: Same when all the ends are 0 except for the first: 
image

fraincs: Good catch, going back to the drawing board.

vicky-comeau: not sure why it wasn't like that to begin with... maybe because the tooltip used to be instead the component and you couldn't have overflow hidden. but if it's easy to do, do it that way.

fraincs: Yes that was not possible, I'm going to do this in JS.

vicky-comeau (approved):

PR #726, opened 2024-02-26
Fix: [IGL-128] Tabs should use hopper icons for Workleap brand

vicky-comeau:


alexasselin008 (approved):

packages/Tabs/src/tabs.scss
```diff
@@ -1,8 +1,11 @@
 @use '~@igloo-ui/tokens/dist/base10/variables' as tokens;
 @use '~@igloo-ui/tokens/dist/fonts';
 
+@layer hopper-design-system, hopper-icons, igloo;
```
fraincs, line 4: Wouldn't this be "global"?

Issue #731, opened 2024-02-26
[Bug]: There's a missing flex-shrink: 0 on the icons of the ActionMenu options


CarolineSabourin-Jean:
### What happened?

There's a missing `flex-shrink: 0` on the icons of the ActionMenu options. Here's an exemple from storybook without the flex-shrink:
![image](https://github.com/gsoft-inc/ov-igloo-ui/assets/25105785/6826deaf-8cd8-4013-befd-c654971a31a8)
 
Here is with the flex-shrink:
![image](https://github.com/gsoft-inc/ov-igloo-ui/assets/25105785/021cd3ca-262a-4a05-ba07-9dcffcbe67e0)

To fix has to be done for both brands :)

Thanks!

### Component

ActionMenu

### Component Version

1.5.1

PR #733, opened 2024-02-27
Hotfix/731

fraincs:
Fix for #731.

Icons in a List could shrink, it was unintended as icons are not made to be scaled.

### DIFF
--- a/packages/List/src/list-item.scss
+++ b/packages/List/src/list-item.scss
@@ -290,7 +290,7 @@
         }
     }
 
-    .ids-visual-identifier__icon {
+    .ids-visual-identifier__visual {
         flex-shrink: 0;
     }
 
--- a/packages/List/src/list-item.scss
+++ b/packages/List/src/list-item.scss
@@ -290,6 +290,10 @@
         }
     }
 
+    .ids-visual-identifier__icon {
+        flex-shrink: 0;
+    }
+
     .ids-list-item--loading .ids-list-item__thumbnail {
         animation: loading-pulse-light ease-in-out 1.5s infinite both;
         border-radius: var(--ids-list-item-loading-thumbnail-border-radius);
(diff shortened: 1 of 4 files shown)


packages/List/src/list-item.scss
```diff
@@ -290,6 +290,10 @@
         }
     }
 
+    .ids-visual-identifier__icon {
```
vicky-comeau, line 293: I'd probably use the class: ids-list-item__visual since it's not just the icon that should have this, the avatar and color should too

vicky-comeau (dismissed):

alexasselin008:
It seems like a odd place to put the fix? Adding more in thread of this discussion

packages/List/src/list-item.scss
```diff
@@ -290,6 +290,10 @@
         }
     }
 
+    .ids-visual-identifier__visual {
```
alexasselin008, line 293: It seems like a odd place to put the fix?

the list item has the problem, so you use css to target the styling of the visual-identifier component.
Why not handle this at the source? either in the visual identifier component, or in the icons directly?

vicky-comeau: I'm not sure I agree with this. For one, maybe it's not always the case, plus you never know what they'll use for icons. We don't have slots like other design systems.
Maybe it could be handled with Visual Identifier, but currently there is no css there. Plus it's only used inside lists and tags I believe. It's an internal component.

fraincs: It makes sense to never shrink a visual identifier.

shared/components/src/visual-identifier/visual-identifier.scss
```diff
@@ -0,0 +1,3 @@
+.ids-visual-identifier__visual {
```
vicky-comeau, line 1: the class here is just ids-visual-identifier

vicky-comeau (approved):

PR #734, opened 2024-02-27
Fix: [IGL-127] TagPicker update

vicky-comeau:


packages/Color/src/Color.tsx
```diff
@@ -33,11 +35,19 @@ const Color: React.FunctionComponent<ColorProps> = ({
     className,
     color,
     dataTest,
+    isSquare = false,
```
alexasselin008, line 38: A color is round by default with the workleap theme.
A color is square by default with the igloo theme.

There is a props isSquare that is false by default, and probably do nothing in the igloo theme?
should this default value be ajusted by the current brand? and isSquare=false with the igloo theme should be round?

vicky-comeau: I'm a bit confused.. but yes it does nothing for the Igloo brand.
don't think we can edit defaults based on brand. 
It would be pointless to do it after since it wouldn't be shown in the storybook props table.
The igloo theme doesn't have a round Color. it's always square.

packages/Color/src/Color.tsx
```diff
@@ -12,6 +12,8 @@ export interface CustomCSSProps extends React.CSSProperties {
 }
 
 export interface ColorProps {
+    /** The appearance of the color */
```
alexasselin008, line 15: Can we add a small comment saying this props is only usable in the brand workleap

packages/Color/src/Color.tsx
```diff
@@ -90,9 +100,10 @@ const Color: React.FunctionComponent<ColorProps> = ({
         <div
             className={classes}
             data-test={dataTest}
-            style={styleProps}
         >
-            {children || initialsFromName}
+            <div className="ids-color__inner"
```
alexasselin008, line 104: weird indentation

packages/TagPicker/src/TagPicker.tsx
```diff
@@ -126,6 +138,35 @@ ref: React.Ref<HTMLInputElement>) => {
...
+                label: result.text,
+                description: result.subtext,
+                type: "list",
+                color: result.color,
+                icon: result.icon,
+                src: result.src
+            };
+
+            if (result.type === "member") {
+                listResultItem = {
```
alexasselin008, line 154: listResultItem.type = "member";
listResultItem.manager = result.isManager;

simpler than duplicate everything?

vicky-comeau: ya you have a point there :P

alexasselin008 (approved):

PR #736, opened 2024-02-27
Fix: [IGL-131] Button should be square when only showing a. icon on m…

vicky-comeau:
…obile

### DIFF
--- a/packages/Button/src/button.scss
+++ b/packages/Button/src/button.scss
@@ -15,6 +15,8 @@
         --ids-btn-line-height-small: normal;
         --ids-btn-height: #{tokens.$space-6};
         --ids-btn-height-small: #{tokens.$space-5};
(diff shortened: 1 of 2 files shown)


alexasselin008 (dismissed):

packages/Button/src/button.scss
```diff
@@ -497,7 +501,12 @@
       }
 
       .ids-btn--mobile {
-        width: auto;
+        padding: 0;
+        width: var(--ids-btn-width);
```
fraincs, line 505: You could have made it with `aspect-ratio: 1/1;` and not having to declare 4 variables.

vicky-comeau: hmm you have a point. I'll try this, thanks

fraincs (approved):

PR #737, opened 2024-02-27
fix for 729 - buttons should be at the bottom when in a Modal

fraincs:
When in mobile buttons should be at the bottom per design discussions. This PR aims at fixing this.

### DIFF
--- a/packages/Modal/src/modal.scss
+++ b/packages/Modal/src/modal.scss
@@ -292,7 +292,7 @@
     .ids-modal__content {
         display: flex;
         flex-direction: column;
-        flex-grow: 1;
+        flex: 1 1 auto;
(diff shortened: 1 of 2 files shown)


packages/Modal/src/modal.scss
```diff
@@ -300,12 +305,23 @@
         }
     }
 
+    .ids-modal__carousel {
+        display: flex;
+        flex-direction: column;
+        flex-grow: 1;
```
vicky-comeau, line 311: I feel like we should always set all three, whether you separate the props or not.
I like to do flex: 1 1 auto; to ensure each browser uses the same.

vicky-comeau (approved):

PR #750, opened 2024-03-01
Fix: [IGL-136] Localization

vicky-comeau:


packages/AreaChart/src/AreaChart.tsx
```diff
@@ -139,7 +143,19 @@ const AreaChart: React.FunctionComponent<AreaChartProps> = ({
     const DEFAULT_SKELETON_WIDTH = 24;
```
alexasselin008, line 143: The constants could be put outside of the component. This way they won't get initialized on every render

packages/AreaChart/src/AreaChart.tsx
```diff
@@ -139,7 +143,19 @@ const AreaChart: React.FunctionComponent<AreaChartProps> = ({
...
     const DEFAULT_SKELETON_HEIGHT = 8;
 
-    Settings.defaultLocale = locale;
+    const localeMapping: Record<string, string> = {
+        "en": "en-US",
+        "fr": "fr-CA"
+    };
+    const providerLocale = useLocale();
+    let actualLocale = providerLocale.locale;
```
alexasselin008, line 151: Can this synthax be used? we do this a lot in orbit

```
    // in the props deconstruction
    locale: localeProp,
...
   const providerLocale = useLocale();
   const locale = localeMapping[locale] ?? providerLocale 

```

packages/Datepicker/src/Datepicker.tsx
```diff
@@ -271,19 +275,17 @@ const Datepicker: React.FunctionComponent<DatepickerProps> = ({
...
-                content={calendar}
-                size="medium"
-                className="ids-datepicker__dropdown"
-                dataTest={dataTest}
-                {...rest}
-            >
-                <Input {...inputProps} />
-            </Dropdown>
-        </I18nProvider>
+        <Dropdown
```
alexasselin008, line 278: hmmmmmmmmmmm are you 100% sure that removing the i18nProvider here is ok? 

Even if i know it's the same provider that is set in our context, the way igloo builds with rollup, they inline the code of 3rd party. So the version of our i18nProvider could be different from the one from the dropdown. 

i think it's safer here to keep it. 

Not going to be the case in hopper however

packages/Metric/src/Score.tsx
```diff
@@ -64,7 +67,9 @@ const Score: React.FunctionComponent<ScoreProps> = ({
             })}
         />
     );
-    const postFix = absoluteValue === 1 ? " pt" : " pts";
+    const postFix = absoluteValue === 1 ?
```
alexasselin008, line 70: really really not important, but i'm like 90% sure that in english, 0 or multiple is plural, and 1 is singular. 
in french, it's 0 or 1 singular, multiple is plural.

packages/AreaChart/src/AreaChart.tsx
vicky-comeau: I didn't do that because the user can set either en of en-US for example.
Unfortunately we supported just en before, but in the component, I want to change it to use en-US.
but I want to support them setting en-US as well (which they may have been doing anyway... Hard to tell in OV's code to be honest..

packages/Datepicker/src/Datepicker.tsx
vicky-comeau: but we want it to use our provider. That's why I added the calendar stuff to our provider.
I tested it and it was working.

packages/Metric/src/Score.tsx
vicky-comeau: ugh that's unfortunate.. :( 
if the messages were working properly, we are able to put plural rules inside the intl file :(

packages/Modal/src/Modal.tsx
```diff
@@ -219,7 +231,7 @@ const Modal: React.FunctionComponent<ModalProps> = (props: ModalProps) => {
                                         className="ids-modal__close"
                                         onClick={onClose}
                                         appearance={{ type: "ghost", variant: "secondary" }}
-                                        aria-label={closeBtnAriaLabel}
+                                        aria-label={stringFormatter.format("close")}
```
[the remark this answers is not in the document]
vicky-comeau: I suppose we could

packages/Provider/src/Provider.tsx
```diff
@@ -1,6 +1,13 @@
 import * as React from "react";
 
-import { I18nProvider, type I18nProviderProps, useLocale, useLocalizedStringFormatter } from "react-aria";
+import { I18nProvider, 
+    type I18nProviderProps, 
+    useLocale, 
+    useLocalizedStringFormatter, 
+    useCalendar,
```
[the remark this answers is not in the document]
vicky-comeau: ok :(
I put them here because it was the only way I could get it to work with no errors, but I guess if you say to put the provider back, then it would probably work that way.

alexasselin008 (approved):

PR #751, opened 2024-03-01
Feat: [IGL-137] Added tertiary action to Modal

vicky-comeau:


packages/Modal/src/Modal.stories.tsx
```diff
@@ -249,6 +250,14 @@ export const WithActions = {
               Cancel
             </Button>
           }
+          tertiaryAction={
+            <Hyperlink
```
alexasselin008, line 254: You should put a ghost danger button, it's not a hopper design but it's still better than an hyperlink

alexasselin008 (approved):

packages/Modal/src/Modal.stories.tsx
vicky-comeau: ok I'll switch it

alexasselin008 (approved):

PR #765, opened 2024-03-08
Centered default and primary text - added inactive appearance in Tag

fraincs:
Tag

- Compensated centering of text in Tags using ABC Favorit Mono
- Added Inactive apperance to Tag

### DIFF
--- a/packages/Tag/src/tag.scss
+++ b/packages/Tag/src/tag.scss
@@ -47,12 +47,40 @@
         --ids-tag-background-primary: #{tokens.$electric-blue-500};
         --ids-tag-color-primary: #{tokens.$samoyed};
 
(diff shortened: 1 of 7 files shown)


packages/Tag/src/tag.scss
```diff
@@ -128,13 +129,17 @@
         --ids-tag-background-primary: var(--hop-primary-surface-strong);
         --ids-tag-color-primary: var(--hop-primary-text-strong);
 
+        /* Inactive */
```
vicky-comeau, line 132: I think we should at least put something for old brand. Even if it's the default tag look... or pick similar colors to inactive. because right now it looks like this:
image

vicky-comeau (approved):

PR #788, opened 2024-04-12
Update Workflows

alexasselin008:
Modified chromatic.yml workflow
- Changed trigger for suggested chromatic triggers
- updated actions/setup-node actions/cache and actions/checkout to v3
- now use the .nvmrc file to dictate node version
- added turbosnap
- Now only build shared and components, skipping the docs. This improves the workflow executionby like 2mins

Modified publish.yml
- now use the .nvmrc file to dictate node version
- cache@v3's restore key had a typo inside it (runner.ox instead of .os)
- Since the publish action is done on main, and main is always buildable / linted properly, i've removed the build / lint / test step. This improves the workflow execution by at least 5 minutes

Modified ci.yml
- now use the .nvmrc file to dictate node version
- cache@v3's restore key had a typo inside it (runner.ox instead of .os)
- The step build_doc was intended to only build the documentation website + storybook, but was rebuilding all the components. I merged build_and_test_components and build_doc to remove the build duplication


.github/workflows/publish.yml
```diff
@@ -5,143 +5,44 @@ on:
...
+              with:
+                path: node_modules
+                key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+                restore-keys: |
+                    ${{ runner.os }}-yarn-
+
+            - name: Install Dependencies
+              if: steps.cache-dependencies.outputs.cache-hit != 'true'
+              run: |
+                yarn install --force --non-interactive
```
alexasselin008, line 37: Is there a reason why we update dependencies in the publish workflow? i should ask Franck

alexasselin008 (commented):
Chromatic seems to  build the documentation website. We should optimize that, this is not needed

.github/workflows/publish.yml
franckgaudin: I can't remember how far back. We can try to remove it

franckgaudin (approved):

Issue #787, opened 2024-04-11
[Bug]: Metric component - icon and box-shadow fix


JeffreyLYT:
### Contact Details

jeffrey.lam@gsoft.com

### What happened?

There seems to be a mismtach between the component design and igloo
Notably, the icon should be bigger and there should be a slight box-shadow to the metric tile

https://www.figma.com/file/KXrVZlFqD8mmlxJJhL5Jhn/%E2%9D%84%EF%B8%8F-Igloo-Design-System?type=design&node-id=4051-2&mode=dev

Prod:
![image](https://github.com/gsoft-inc/ov-igloo-ui/assets/43586477/0dcbe294-3e42-4be6-b61c-ff9461ad7fa9)
Design:
![image](https://github.com/gsoft-inc/ov-igloo-ui/assets/43586477/a9dcdc83-c5be-4a02-b177-848ea96a59d7)

### Component

Metric

### Component Version

0.3.3

### Which browsers are you seeing the problem on?

Chrome

fraincs:
The box shadow applied is the raised one, like in Figma, this won't be fixed. As for the arrow I am fixing this.

PR #798, opened 2024-04-16
usage of hopper icon in metric score

fraincs:
Fix for issue : https://github.com/gsoft-inc/ov-igloo-ui/issues/787

### DIFF
--- a/packages/Metric/src/Metric.tsx
+++ b/packages/Metric/src/Metric.tsx
@@ -3,9 +3,10 @@ import cx from "classnames";
 import { useButton, type AriaButtonProps, FocusRing } from "react-aria";
 
 import HelpSolid from "@igloo-ui/icons/dist/HelpSolid";
-import Help from "@igloo-ui/icons/dist/Help";
 import Alignment from "@igloo-ui/icons/dist/Alignment";
 import Checkmark from "@igloo-ui/icons/dist/Checkmark";
+
+import { QuestionIcon } from "@hopper-ui/icons-react16";
 
 import Tooltip from "@igloo-ui/tooltip";
 import Score from "./Score";
@@ -109,7 +110,7 @@ const Metric: React.FunctionComponent<MetricProps> = ({
                         {tooltip && appearance !== "selected" && (
                             <Tooltip content={tooltip}>
                                 {isWorkleap ? (
-                                    <Help size="small" />) : <HelpSolid size="small" />}
+                                    <QuestionIcon size="sm" />) : <HelpSolid size="small" />}
                             </Tooltip>
                         )}
                         {type !== "subMetric" && appearance === "selected" && (
--- a/packages/Metric/src/Metric.tsx
+++ b/packages/Metric/src/Metric.tsx
@@ -54,6 +54,8 @@ const Metric: React.FunctionComponent<MetricProps> = ({
         return document.documentElement.getAttribute("data-brand") ?? "igloo";
     };
 
+    const isWorkleap = getBrand() === "workleap";
+
     const { buttonProps } = useButton(rest, btnRef);
 
     const classes = cx("ids-metric", className, {
@@ -106,7 +108,7 @@ const Metric: React.FunctionComponent<MetricProps> = ({
                     >
                         {tooltip && appearance !== "selected" && (
                             <Tooltip content={tooltip}>
-                                {getBrand() === "workleap" ? (
+                                {isWorkleap ? (
                                     <Help size="small" />) : <HelpSolid size="small" />}
                             </Tooltip>
                         )}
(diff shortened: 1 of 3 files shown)


packages/Metric/src/Metric.tsx
```diff
@@ -106,7 +108,7 @@ const Metric: React.FunctionComponent<MetricProps> = ({
                     >
                         {tooltip && appearance !== "selected" && (
                             <Tooltip content={tooltip}>
-                                {getBrand() === "workleap" ? (
+                                {isWorkleap ? (
                                     <Help size="small" />) : <HelpSolid size="small" />}
```
vicky-comeau, line 112: workleap icon should be "sm"

packages/Metric/src/Score.tsx
```diff
@@ -53,28 +79,48 @@ const Score: React.FunctionComponent<ScoreProps> = ({
...
     }
+
     const arrow = isNegative ? (
-        <ArrowDown
-            size={arrowSize}
-            className={cx("ids-score__arrow", "ids-score__arrow--negative", {
-                "ids-score__arrow--selected": isSelected
-            })}
-        />
+        isWorkleap ? (
```
vicky-comeau, line 84: way too many ternary operations going on here.
I think it can be simplified or split up.
Since you have a size adapter, I'd remove this isWorkleap condition from here and put it back to how is was.
above you can create two new consts.
one for the icons classes.
The other would be the icon's name.

packages/Metric/src/Score.tsx
```diff
@@ -3,6 +3,8 @@ import cx from "classnames";
 
 import ArrowUp from "@igloo-ui/icons/dist/ArrowUp";
 import ArrowDown from "@igloo-ui/icons/dist/ArrowDown";
+import { ArrowUpIcon, ArrowDownIcon } from "@hopper-ui/icons-react16";
```
vicky-comeau, line 6: you should be able to import `type IconProps` here.

packages/Metric/src/Score.tsx
```diff
@@ -36,6 +38,30 @@ const Score: React.FunctionComponent<ScoreProps> = ({
...
     const stringFormatter = useLocalizedStringFormatter(intlMessages);
     const { locale } = useLocale();
+
+    const getBrand = (): string => {
+        return document.documentElement.getAttribute("data-brand") ?? "igloo";
+    };
+
+    const isWorkleap = getBrand() === "workleap";
+
+    const sizeAdapter = (size: "small" | "medium" | "large"): "sm" | "md" | "lg" => {
```
vicky-comeau, line 48:
```suggestion
    const sizeAdapter = (size: "small" | "medium" | "large"): IconProps["size"] => {
```

packages/Metric/src/Metric.tsx
fraincs: None of these are Workleap icon though

fraincs: I only changed the conditional statement while at it. Help and helpsolid are both Igloo icons

packages/Metric/src/Score.tsx
```diff
@@ -36,6 +38,35 @@ const Score: React.FunctionComponent<ScoreProps> = ({
...
+    const arrowPositiveClass = cx("ids-score__arrow", "ids-score__arrow--positive", {
+        "ids-score__arrow--selected": isSelected
+    });
+
+    const arrowNegativeClass = cx("ids-score__arrow", "ids-score__arrow--negative", {
+        "ids-score__arrow--selected": isSelected
+    });
+
+    const ArrowUpIconElement = isWorkleap ?
+        <ArrowUpIcon className={arrowPositiveClass} size={sizeMap[arrowSize] as IconProps["size"]} /> :
```
vicky-comeau, line 63: this does work but I meant more
`const ArrowUpIconComponent = isWorkleap ? ArrowUpIcon : ArrowUp;`
Then this could be used directly in the component.

packages/Metric/src/Score.tsx
```diff
@@ -36,6 +38,35 @@ const Score: React.FunctionComponent<ScoreProps> = ({
...
     const stringFormatter = useLocalizedStringFormatter(intlMessages);
     const { locale } = useLocale();
+
+    const getBrand = (): string => {
+        return document.documentElement.getAttribute("data-brand") ?? "igloo";
+    };
+
+    const isWorkleap = getBrand() === "workleap";
+
+    const sizeMap = {
```
alexasselin008, line 48: this object can be outside the component, so it doesn't get created everytime

alexasselin008: you should also add " as const" at the end, so you won't have to cast it with IconProps["size"] later

packages/Metric/src/Score.tsx
```diff
@@ -36,6 +44,29 @@ const Score: React.FunctionComponent<ScoreProps> = ({
 }: ScoreProps) => {
     const stringFormatter = useLocalizedStringFormatter(intlMessages);
     const { locale } = useLocale();
+
+    const getBrand = (): string => {
```
vicky-comeau, line 48: sorry I didn't notice this before. I usually put this function outside of Score function component.

vicky-comeau (approved):

PR #810, opened 2024-05-10
fix callout horizontal font weight

fraincs:
Horizontal Callout in Workleap theme should have a font-weight of 410, this PR fixes it.


packages/Alert/src/alert.scss
```diff
@@ -51,6 +51,7 @@
         --ids-alert-font-size: var(--hop-body-sm-semibold-font-size);
         --ids-alert-line-height: var(--hop-body-sm-semibold-line-height);
         --ids-alert-title-font-weight: var(--hop-body-sm-semibold-font-weight);
+        --ids-alert-horizonta-title-font-weight: var(--hop-body-sm-font-weight);
```
vicky-comeau, line 54: missing an l

vicky-comeau (approved):

PR #815, opened 2024-05-16
updated hopper tokens version

fraincs:
Updated Orbiter to latest Hopper tokens version (4.0.0)


packages/Button/package.json
vicky-comeau, line 1: hover seems to be wrong for the primary button:
image

vicky-comeau (approved):

PR #817, opened 2024-05-23
Feature/pressed state tokens

fraincs:
- Added pressed states to components that needed it.
- Fixed UI for fields focus style.

### DIFF
--- a/.changeset/old-items-arrive.md
+++ b/.changeset/old-items-arrive.md
@@ -1,0 +1,28 @@
+---
+"@igloo-ui/button": minor
+"@igloo-ui/checkbox": minor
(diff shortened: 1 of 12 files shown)


.changeset/old-items-arrive.md
vicky-comeau, line 1: Possible components missed: 
- ButtonGroup
- DatePicker date
- Disclosure
- Filter
- Input? can't remember if pressed state actually works, maybe not
- list item
- Metric (maybe not as important)
- OptionButton
- Pager (also probably shouldn't have a blue border when clicking)
- Tab (barely a difference but exists in figma)
- Textarea, but same concern as Input
- same with TextEditor

PR #820, opened 2024-05-28
updated Disclosure UI in workleap theme

fraincs:
Updated Disclosure to match the intended look in Hopper.

### DIFF
--- a/packages/Disclosure/src/disclosure.scss
+++ b/packages/Disclosure/src/disclosure.scss
@@ -74,10 +74,10 @@
         --ids-disclosure-header-line-height: var(--hop-body-sm-line-height);
         --ids-disclosure-header-padding: var(--hop-space-inset-md);
         --ids-disclosure-header-icon-margin: 0 var(--hop-space-inline-md) 0 0;
-        --ids-disclosure-header-title-font-family: var(--hop-heading-xs-font-family);
-        --ids-disclosure-header-title-font-size: var(--hop-heading-xs-font-size);
-        --ids-disclosure-header-title-font-weight: var(--hop-heading-xs-font-weight);
-        --ids-disclosure-header-title-line-height: var(--hop-heading-xs-line-height);
+        --ids-disclosure-header-title-font-family: var(--hop-heading-lg-font-family);
+        --ids-disclosure-header-title-font-size: var(--hop-heading-lg-font-size);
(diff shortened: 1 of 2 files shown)


packages/Disclosure/src/disclosure.scss
```diff
@@ -213,14 +213,13 @@
     }
 
     .ids-disclosure__header-content {
-        display: block;
         margin: var(--ids-disclosure-header-content-margin);
+        display: flex;
```
vicky-comeau, line 217: this doesn't quite work because the description is supposed to be under the title and it's not doing that:
image
you can see that the description moves up slightly because the top margin is removed.

packages/Disclosure/src/disclosure.scss
```diff
@@ -213,14 +213,13 @@
...
+        flex: 1 1 auto;
 
         /* stylelint-disable-next-line media-query-no-invalid */
         @media (width >= #{tokens.$breakpoints-sm}) {
             align-items: baseline;
-            display: flex;
```
vicky-comeau, line 222: this should stay here and only flex: 1 1 auto should be moved out of the breakpoint.

packages/Disclosure/src/disclosure.scss
```diff
@@ -213,14 +213,13 @@
     }
 
     .ids-disclosure__header-content {
-        display: block;
```
vicky-comeau, line 216: keep block

packages/Disclosure/src/disclosure.scss
```diff
@@ -213,14 +213,13 @@
...
         margin: var(--ids-disclosure-header-content-margin);
+        display: flex;
+        flex: 1 1 auto;
```
vicky-comeau, line 218: this is all we need for the fix.

packages/Disclosure/src/disclosure.scss
```diff
@@ -213,14 +213,14 @@
     }
 
     .ids-disclosure__header-content {
-        display: block;
         margin: var(--ids-disclosure-header-content-margin);
+        display: flex;
```
vicky-comeau, line 217: having display flex here will still not allow wrapping like we want, that's why I think display block is better here.

vicky-comeau (approved):

PR #827, opened 2024-06-03
Fix Alert UI to match Hopper style

fraincs:
- Padding is now smaller when in Workleap theme.
- Button is not ghost in horizontal appearance anymore in Workleap theme.
- In mobile, the items stack instead of being horizontal.


packages/Alert/src/alert.scss
```diff
@@ -119,6 +119,7 @@
 
         &--inline,
         &--horizontal {
+            padding: 0.5rem;
```
vicky-comeau, line 122: this needs to be a variable. it's probably different between igloo and workleap brand

fraincs: Totally, this was a placeholder.

vicky-comeau (approved):

PR #845, opened 2024-08-28
Feat: [DS-288] Added support for action link & closable horizontal ap…

vicky-comeau:
…pearance

### DIFF
--- a/packages/Avatar/src/__snapshots__/Avatar.test.tsx.snap
+++ b/packages/Avatar/src/__snapshots__/Avatar.test.tsx.snap
@@ -4,6 +4,7 @@ exports[`Avatar It should render a snapshot 1`] = `
 <DocumentFragment>
   <div
     class="ids-avatar ids-avatar--medium"
+    data-private="true"
   >
     <img
       alt="Avatar"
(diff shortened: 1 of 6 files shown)


packages/Avatar/src/__snapshots__/Avatar.test.tsx.snap
```diff
@@ -4,6 +4,7 @@ exports[`Avatar It should render a snapshot 1`] = `
 <DocumentFragment>
   <div
     class="ids-avatar ids-avatar--medium"
+    data-private="true"
```
fraincs, line 7: is that from another PR?

packages/Dropdown/src/__snapshots__/Dropdown.test.tsx.snap
```diff
@@ -2,7 +2,9 @@
 
 exports[`Dropdown It should render without error and a snapshot 1`] = `
 <body>
-  <div>
+  <div
+    data-floating-ui-inert=""
```
fraincs, line 6: are the empty quotes necessary?

packages/Avatar/src/__snapshots__/Avatar.test.tsx.snap
vicky-comeau: I didn't refresh the snapshot last time. I don't remember it complaining.. weird. so you can ignore this.

packages/Dropdown/src/__snapshots__/Dropdown.test.tsx.snap
vicky-comeau: it's a snapshot. auto generated

fraincs (approved):

PR #849, opened 2024-09-16
updated to latest hopper tokens/icons

fraincs:
updated to latest hopper tokens and icons


.changeset/slimy-moose-float.md
vicky-comeau, line 1: were there not any hopper rich icons used?

vicky-comeau: oh wait, even if there was, there's not actually a change that we need to make, is there?

fraincs: The richIcon variables are not used, this is what was changed.

vicky-comeau (dismissed):

PR #852, opened 2024-09-19
Updated typings

fraincs:
- Added the possibility to use a reactNode in Toaster message.
- Added the possibility to use a reactNode in ListItem label.

`onClick` typings have been extended from `void` to `MouseEventHandler<HTMLButtonElement`.
- ActionMenuOption
- Button
- IconButton
- Filter
- Stepper

### DIFF
--- a/packages/Alert/src/Alert.tsx
+++ b/packages/Alert/src/Alert.tsx
@@ -30,7 +30,7 @@ export type Appearance = "card" | "inline" | "horizontal";
 
 export interface AlertButton {
     label: React.ReactNode;
-    onClick: () => void;
+    onClick: (() => void) | React.MouseEventHandler<HTMLButtonElement>;
 }
 
 export interface AlertProps extends Omit<React.ComponentProps<"div">, "title"> {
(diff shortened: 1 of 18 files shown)


packages/Alert/src/Alert.tsx
```diff
@@ -30,7 +30,7 @@ export type Appearance = "card" | "inline" | "horizontal";
 
 export interface AlertButton {
     label: React.ReactNode;
-    onClick: () => void;
+    onClick?: (e?: React.SyntheticEvent) => void;
```
victortrinh2, line 33: Est-ce qu'on voulait rendre le onClick nullable?

packages/Toaster/src/Toaster.stories.tsx
```diff
@@ -115,4 +115,17 @@ export const ErrorToast = () => {
         <Toaster />
     </ChromaticWrapper>
   );
-};
\ No newline at end of file
+};
+
+
+export const NodeMessageSuccessToast = () => {
+
```
victortrinh2, line 122: [Super nit] extra spaces in line 120 and 122

victortrinh2 (approved):

PR #880, opened 2025-07-22
EP-6633 Restore accidentally removed CSS variables

PatrickDeVries:
I accidentally removed a few CSS variables when cleaning up the old brand styles. I naively replaced the old styles with the rebrand styles, but a few CSS vars were only defined in the first block so this removed them. This restores them. To find the issues I went through and made sure each CSS var declaration block has the same or more vars as the original block did.

### DIFF
--- a/packages/Disclosure/src/disclosure.scss
+++ b/packages/Disclosure/src/disclosure.scss
@@ -39,6 +39,8 @@
         /* Content */
         --ids-disclosure-content-border: 0.0625rem solid var(--hop-neutral-border-weak);
 
+        /* === Low Contrast === */
+
         /* Header/Button */
         --ids-disclosure-low-contrast-header-padding: 0;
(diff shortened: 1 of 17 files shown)


packages/Disclosure/src/disclosure.scss
```diff
@@ -39,6 +39,8 @@
         /* Content */
         --ids-disclosure-content-border: 0.0625rem solid var(--hop-neutral-border-weak);
 
+        /* === Low Contrast === */
```
victortrinh2, line 42: Is this needed?

PatrickDeVries: No it doesn't affect anything, I just restored it since it was there before. The vars below that point are all for the low contrast option though so it isn't wrong. Not sure how we want to lean for comments, should I just remove it?

victortrinh2: Nono it's all good!

victortrinh2 (approved):

